PR #2, opened 2025-05-28
tests

ehmry:
A basic test driver in Tcl. Made with some QEMU and mkfs code copied from Nixpkgs.

I'm fine with reorganising the layout of these files.

The _modules/boot/init: use textClosureList for PID 1 arguments_ commit isn't necessary but it came with when I extracted this series from my Synit branch.


aanderse (changes requested):
this is wonderful!

> I'm fine with reorganising the layout of these files.

unless you have any specific ideas i think we can merge without for now

> The _modules/boot/init: use textClosureList for PID 1 arguments_ commit isn't necessary but it came with when I extracted this series from my Synit branch.

looks like we can cleanly drop that commit from this PR - i would like to do that because the option is added, but not utilized yet

PR #4, opened 2025-06-08
add synit modules

ehmry:
This is the big Synit PR.

### DIFF
--- a/modules/boot/default.nix
+++ b/modules/boot/default.nix
@@ -1,4 +1,4 @@
-{ config, pkgs, ... }:
+{ config, lib, pkgs, ... }:
 {
   imports = [
     ./initrd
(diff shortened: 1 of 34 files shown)


aanderse (approved):
looked over all commits in detail except for the last one as i'm not familiar with `synit` yet... awesome!

- i booted my system into this and can confirm no regressions
- i would love to try `mevd` out at some point, i have never used it

thanks!

---

btw... at some point i think we need to look into tweaking the `boot.kernelParams` a bit so we can export a `/kernel-params` file in the `system.topLevel` derivation... this would improve boot loader capabilities

ehmry:
Excellent. I agree that we are weak on bootloaders right now.

Mdevd works, but in the current state it just inserts modules and creates symlinks in `/dev/disk/by-*`. I've booted my laptop using the [mdev-like-a-boss](https://github.com/fff7d1bc/mdev-like-a-boss) config but I haven't examined it closely yet.

PR #5, opened 2025-06-11
more synit documentation and mdevd 

ehmry:


modules/services/dhcpcd/default.nix
```diff
@@ -167,7 +175,22 @@ in
...
-      command = "${pkgs.dhcpcd}/bin/dhcpcd -f ${cfg.configFile}";
+      command = "${lib.getExe cfg.package} -f ${cfg.configFile}";
+    };
+
+    synit.daemons.dhcpcd = let
+      # TODO: This should be handled by tmpfiles.
+      script = pkgs.execline.passthru.writeScript "dhcpcd-dirs.el" "-s0" ''
+        foreground { ln -s /run /var/run }
+        foreground { install --directory --owner=dhcpcd /var/db/dhcpcd }
+        foreground { install --directory --owner=dhcpcd --group=dhcpcd /var/lib/dhcpcd }
```
aanderse, line 186: TODO: switch to `tmpfiles` in the short term - either `finit`, `systemd`, or even [pawprint](https://github.com/eweOS/pawprint)... and then eventually consider a better interface

aanderse (approved):
awesome! thanks @ehmry

PR #6, opened 2025-06-14
modules/mdevd: add support for finit

aanderse:
currently this doesn't boot into a graphical (`wayland`) environment because of `libinput`s dependency on `udev`

### DIFF
--- a/modules/services/mdevd/default.nix
+++ b/modules/services/mdevd/default.nix
@@ -111,6 +111,37 @@ in
         }
     ];
 
+    finit.services.mdevd = {
+      description = "device event daemon (mdevd)";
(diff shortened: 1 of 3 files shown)


aanderse:
hey @alyssais 👋 

i was wondering if you had any useful information for me regarding running `wayland` based software on an `mdevd` system? i am a bit stuck because i don't know what to do with `niri` which depends on `libinput`...

i looked at some of [your work](https://spectrum-os.org/git/spectrum/tree/host/rootfs/default.nix) but wasn't exactly sure what i could do...

specifically i am asking - do you have any experience with/advice for building and running `libinput` from `nixpkgs`? i am a complete `mdevd` noob so pardon my ignorance. i really appreciate any advice you can give, thanks! 🙇‍♂️

alyssais:
Is the problem that libinput depends on libudev?  I use an overlay (in host/rootfs/default.nix) that links everything with libudev-zero, which is an implementation of the udev API that doesn't require udevd. Perhaps that's what you need?

PR #8, opened 2025-06-25
modules/initrd: import ZFS pools with fileSystemImportCommands

ehmry:
Inject ZFS commands into the initrd init script from the zfs module instead of explicitly within the initrd module.

### DIFF
--- a/modules/filesystems/zfs.nix
+++ b/modules/filesystems/zfs.nix
@@ -24,9 +24,36 @@
         default = [ pkgs.zfs ];
       };
     };
+
+    boot.zfs = {
+      importPools = lib.mkOption {
+        type = with lib.types; listOf str;
+        description = ''
+          List of ZFS pools to import at boot.
+          Defaults to the pools necessary for booting.
+        '';
+        example = [ "jug" "bucket" ];
+      };
+      loadKeys = lib.mkOption {
+        type = with lib.types; listOf str;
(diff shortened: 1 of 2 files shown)


modules/filesystems/zfs.nix
```diff
@@ -24,9 +24,35 @@
         default = [ pkgs.zfs ];
       };
     };
+
+    boot.zfs = {
+      importPools = lib.mkOption {
```
aanderse, line 29:
```suggestion
      importPools = lib.mkOption {
        type = with lib.types; listOf str;
```

aanderse (changes requested):
overall looks good, along with the other PR

i will tests these two tonight and we can merge 👍

thanks!

aanderse:
btw did you have any ideas for building an initrd with something other than bash scripts and nix?

last year i built one with `dinit` and it was pretty nice... would `synit` work well in an initrd?

i am not really interested in rebuilding the initrd code yet but i am looking for some ideas when i eventually get around to it

ehmry:
I had an initrd init that was written in execline instead of bash/dash but it was still calling dash to run some snippets that I hadn't converted and I suspect it would be marginally slower anyway.

I've thought about running `synit-pid1` and `syndicate-server` in the initrd but I don't understand how `switch_root` would work.

PR #12, opened 2025-07-10
Synit stuff

ehmry:
A pile of commits that adds daemons and fixes for Synit.


aanderse (approved):
thanks @ehmry!

i especially appreciate the pam related commit - this code was copy+paste from nixos so it is really nice to simplify here :rocket:

PR #13, opened 2025-07-11
overlays/without-udev: overlay for running without udev

ehmry:


aanderse (approved):
awesome! looks good to me, similar to what i used when testing `mdevd` :+1:

PR #14, opened 2025-07-24
synit: mdevd, sysctl, modprobe, fuse, wpa_supplicant, networking

ehmry:
Synit service additions and improvements, and a dedicated FUSE module.


ehmry:
Now with wpa_supplicant (Synit-only) and a rewrite of the networking scripts (https://git.syndicate-lang.org/synit/synit-network-utils).

PR #16, opened 2025-08-01
WiP! static addresses and routes

ehmry:
This is a proposal for a static addresses and routes interface.

Instead of an attrset by device it would be
```nix
{
  networking.ipv4 = {
    addresses = [ ];
    routes = [ ];
  };
  networking.ipv6 = {
    addresses = [ ];
    routes = [ ];
  };
}
```

The examples in the module show the basic attributes for addresses and routes that I think are necessary to start with.

### DIFF
--- a/modules/networking/default.nix
+++ b/modules/networking/default.nix
@@ -1,6 +1,61 @@
 { config, pkgs, lib, ... }:
 let
   cfg = config.networking;
+
+  deviceOpt = lib.mkOption {
+    type = lib.types.str;
+    description = "Network attachment device.";
+    example = "eth0";
+  };
+
+  prefixLengthOpt = lib.mkOption {
+    type = lib.types.int;
+    description = "Network prefix length.";
+  };
+
+  addressesSubmodule = addrLen: {
+    options = {
+      device = deviceOpt;
+
+      local = lib.mkOption {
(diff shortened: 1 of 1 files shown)


aanderse:
the nixos option has addresses as a list... but what you're proposing goes a little bit further and makes even the device part of the list

do we anticipate any issues with that? generally in nixos land lists are bad, attrsets are good

thoughts?

ehmry:
My thinking was that working with a list of all addresses in the system requires less thinking than if they were divided across device attrsets. Dividing by device is easier than dividing by something like globally or locally routeable but it doesn't seem less arbitrary.

Before:
```nix
{ cfg, lib, ... }:
{
  allIpv4Addrs = cfg.networking.devices |> builtins.attrValues |> lib.concatMap (dev: dev.ipv4.addresses);
  eth0Ipv4Addrs = cfg.networking.eth0.ipv4.addresses;
}
```

After:
```nix
{ cfg, lib, ... }:
{
  allIpv4Addrs = cfg.networking.ipv4.addresses;
  eth0Ipv4Addrs = builtins.filter ({ dev, ... }: dev == "eth0" }) cfg.networking.eth0.ipv4.addresses;
}
```

PR #19, opened 2025-08-19
modules/i18n: set locale for PID 1 and in pam_env.conf

ehmry:
- Cleanup the configuration of init and PID 1.
- Add an option to generate `/etc/security/pam_env.conf`.
- Set locale on PID 1 so that it propagates into services.
- Set locale with PAM so that it propagates into logins.

### DIFF
--- a/modules/boot/init.nix
+++ b/modules/boot/init.nix
@@ -40,10 +40,6 @@ in
 
   config = {
 
-    boot.kernelParams = [
-      "init=${cfg.init.script}"
-    ];
-
     boot.init.script = pkgs.writeScript "init" ''
       #!${pkgs.runtimeShell}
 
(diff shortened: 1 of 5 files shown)


aanderse:
thanks @ehmry - i like what you did with `security.pam.env`, thanks :heavy_check_mark:

a few thoughts:
- maybe we should rename `security.pam.env` to `security.pam.environment` to be consistent
- currently my `grub.sh` script entirely ignores `boot.kernelParams` and just hardcodes `init=$toplevel/init` because of cyclic dependencies with `boot.kernelParams` in `system.activation.topLevel` - should we just drop `init=...` from `boot.kernelParams` and hardcode `init=$toplevel/init` where needed (like `qemu` and boot loaders)?

PR #20, opened 2025-08-22
activation refactoring

ghost:
I decide to move the initial configuration activation from before to after the Synit PID 1 starts.

While I was doing that I made the activation script start snipppets in parallel to reduce boot time.

### DIFF
--- a/modules/synit/daemons.nix
+++ b/modules/synit/daemons.nix
@@ -283,6 +283,7 @@ let
       attrs' =
       {
         argv =
+          [ "s6-setlock" "/run/synit/locks/daemon-${name}" ] ++
           # Inject logging.
           optionals attrs.logging.enable (loggerArgs {
 	      reserveStdio = protocol != "none";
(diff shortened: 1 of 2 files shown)


aanderse:
> I hadn't considered removing the activation script completely but I like it the more I think about it. I didn't parallelize the activation script because I noticed that it was slow, it was because I noticed that the dependency information makes parallel jobs possible.

in that case do you mind if we drop the first commit, [run snippets in parallel](https://github.com/aanderse/finix/pull/20/commits/3430fa8449286b23bc4fc96967391651733489b3), from the PR for now? i just think it is a step in the wrong direction given that you agree a service manager in initramfs is desirable

> I would prefer to move Synit into the initramfs stage 1 rather than use another init there. It should be possible.

sorry, i didn't realize that was a possibility with `synit` at this point - great! let's set that as the goal then

> I think we should do something about stage 1 but I'm not ready to think about that before I can do configuration transitions without rebooting.

sure, sounds good

aanderse (approved):
awesome! this is a very nice change - and works fine on my system :+1:

thanks @ehmry

PR #51, opened 2026-04-11
modules/filesystems/btrfs: fix crypto module names

Ollie-Gutierrez:
Use algorithm names that the btrfs kernel module expects (`xxhash64`, `sha256`, `blake2b-256`) instead of generic module names (`xxhash_generic`, `blake2b_generic`, `sha256_generic`), fixed some errors while building for me.

https://github.com/NixOS/nixpkgs/blob/b6a11bfa3ad664644bf6d1991229196d95837a4b/nixos/modules/tasks/filesystems/btrfs.nix#L91-L97


aanderse (approved):
i saw this a while back but hadn't got around to adding to `finix`... thank you so much! 🎉

PR #53, opened 2026-04-13
luks and lvm filesystems

Ollie-Gutierrez:
I haven’t been able to test this properly yet so I would appreciate if someone would give these a go, I’ve tried to emulate the other filesystems modules as much as possible.


Ollie-Gutierrez:
Going to have a proper go at testing these tomorrow

Ollie-Gutierrez:
Hey, just pushing an update on this PR.

I've refactored the modules as there were quite a few issues with the previous implementation, they should now be more correct and better aligned with the other filesystem modules.

However, LUKS support is still not functional. The core issue appears to be an inability to read input from the terminal, which I believe is a consequence of the way Finit handles process I/O, specifically whats discussed in finit-project/finit#120. I've attempted to work around this using pipes and setsuid, but neither approach has resolved it.

I think this might be something that needs to be addressed at a deeper level, happy to hear any thoughts on the right direction here.

Ollie-Gutierrez:
@troglobit curious if you have any insight on this, is there a recommended way to handle interactive input like passphrase prompts in this context, or is this something that would need a change on Finit’s side?

I have seen mention of setting up cryptsetup in the documentation: https://finit-project.github.io/switchroot/?h=switch#example-initramfs-finitconf, however this approach hasn’t worked for me, unless I’m missing something.

troglobit:
> @troglobit curious if you have any insight on this, is there a recommended way to handle interactive input like passphrase prompts in this context, or is this something that would need a change on Finit’s side?

I'm not aware of someone doing something like this before using Finit, so yeah you'd need some basic support in Finit core at least. Which I'm not at all against, I might add.  Patches are most welcome :smiley::+1: and I suggest perhaps looking at how the current rescue mode works for inspiration.

Ollie-Gutierrez:
Everything is now working in my testing, squashed commits to neaten it up since it was a bit of a mess. Just waiting on https://github.com/finit-project/finit/pull/488 to be merged then this will be ready!

PR #54, opened 2026-04-14
services/vnstat: init

chocoblocko9:
Adds a vnstat service. Tested and seems to work just fine!


chocoblocko9:
this is good to merge i think :D

PR #57, opened 2026-04-22
add direct `nixos-rebuild` support for configuration.nix + channels w…

aanderse:
…orkflow

addresses comment from https://github.com/finix-community/finix/issues/29#issuecomment-4277701264 - merge pending additional testing


aanderse:
i found out that `nixos-install` isn't consistent with `nixos-rebuild` and won't look at `nixos-system` from `$NIX_PATH`... which makes this PR less useful... currently working on some documentation on how to install and manage `finix` with channels, more to come

PR #58, opened 2026-04-24
modules/greetd: prevent user session from restarting on updates

aanderse:
currently updating `nixpkgs` and running `nixos-rebuild switch` will cause your graphical session to be killed because `finit` automatically restarts services which need to be restarted

this is great default behaviour... though not always desirable! by providing a stable string for `finit.services.<name>.command` we can avoid restarting services which will kill important services... at the cost of running out of date code

seems like this should possibly be a configurable option? more consideration needed. leaving this here as an example in the meantime.

### DIFF
--- a/modules/services/greetd/default.nix
+++ b/modules/services/greetd/default.nix
@@ -7,8 +7,6 @@
 let
   cfg = config.services.greetd;
   format = pkgs.formats.toml { };
-
-  configFile = format.generate "greetd.toml" cfg.settings;
 in
 {
   options.services.greetd = {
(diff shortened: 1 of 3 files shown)


aanderse:
the next step in this PR is to add an option to etc file generation so a file can be written to the store via `builtins.writeFile` (instead of `pkgs.writeFile`) to completely avoid _any_ reliance on `nixpkgs`... this is the "nuclear" option as it will prohibit any store references at all in the generated file... and provide a path that will _never_ change 

none of this is particularly difficult, but requires careful thought to get the interface right

---

stay tuned, more to come... and in the meantime follow the very good advice nixos provides: reboot your system when you update your `nixpkgs` input!

PR #59, opened 2026-05-02
modules/ly: log to syslog by default

aanderse:
`ly` didn't support log rotation - support has been [added](https://codeberg.org/fairyglade/ly/issues/965) in version `1.5.0`

this is desirable for a few reasons, the most important of which is log rotation

cc @chocoblocko9


aanderse:
tested by @chocoblocko9 - thanks!

PR #61, opened 2026-05-03
feat(limine): add secure boot support

Ollie-Gutierrez:
## Summary

- Add `programs.limine.secureBoot` options for signing the Limine bootloader with `sbctl`
- Support auto-generating and auto-enrolling secure boot keys
- Sign fwupd EFI binaries when secure boot is enabled
- Add assertions to enforce checksum validation, hash mismatch panic, and disabled editor when secure boot is on

### DIFF
--- a/modules/services/fwupd/default.nix
+++ b/modules/services/fwupd/default.nix
@@ -93,6 +93,9 @@ in
       conditions = "service/polkit/ready";
       log = true;
       nohup = true;
+      environment = lib.optionalAttrs (config.programs.limine.secureBoot.enable or false) {
+        FWUPD_EFIAPPDIR = "${cfg.package}/libexec/fwupd/efi";
+      };
     };
 
     finit.tmpfiles.rules = [
--- a/modules/programs/limine/default.nix
+++ b/modules/programs/limine/default.nix
@@ -239,44 +239,125 @@ in
         Force installation even if the safety checks fail, use absolutely only if necessary!
       '';
     };
+
+    secureBoot = {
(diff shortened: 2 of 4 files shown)


modules/services/fwupd/default.nix
```diff
@@ -93,6 +93,9 @@ in
       conditions = "service/polkit/ready";
       log = true;
       nohup = true;
+      environment = lib.optionalAttrs (config.programs.limine.secureBoot.enable or false) {
```
aanderse, line 96: should we add an option to the provider to abstract this instead of under the limine option set?

Ollie-Gutierrez: I'm not too sure, the main reason being I believe limine is the only bootloader that provides secureboot out of the box, the only other option I know of is lanzaboote which is built ontop of systemd-boot.

Ollie-Gutierrez:
@aanderse I think we could get this merged, some more testing would be great but it's been running stable on my laptop a while now so I'm pretty confident it's okay

modules/services/fwupd/default.nix
aanderse: ok, that makes sense

i haven't really thought through the [supportedFeatures](https://github.com/finix-community/finix/blob/9bee5f7dec956440588ca21471284e68954e8161/modules/providers/scheduler/default.nix#L19) aspect of `providers` but i have a feeling that might be a good application here...

let's merge for now, as is, and i will give this some thought. thanks so much for contributing!

aanderse (approved):
amazing!

PR #63, opened 2026-05-08
Fix `services.xserver` unknown drivers section string substitution

deathbymanatee:
Looks like I missed some package string substitution in the `xserver` service; this PR rectifies this mistake and applies the correct string substitution.

### DIFF
--- a/modules/services/xserver/default.nix
+++ b/modules/services/xserver/default.nix
@@ -281,7 +281,7 @@ in
       name:
       let
         driver = lib.attrByPath [ name ] (
-          if pkgs.xorg ? ${"xf8-6video" + name} then { modules = [ pkgs.${"xf86-video" + name} ]; } else null
+          if pkgs ? ${".xf86-video-" + name} then { modules = [ pkgs.${"xf86-video-" + name} ]; } else null
         ) knownVideoDrivers;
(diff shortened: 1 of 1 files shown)


deathbymanatee:
For future reference, when I'm looking to make small commits / fixes related to previous PRs, should I reopen the other PR or create a new one like in this case?

modules/services/xserver/default.nix
```diff
@@ -281,7 +281,7 @@ in
       name:
       let
         driver = lib.attrByPath [ name ] (
-          if pkgs.xorg ? ${"xf8-6video" + name} then { modules = [ pkgs.${"xf86-video" + name} ]; } else null
+          if pkgs ? ${".xf86-video-" + name} then { modules = [ pkgs.${"xf86-video-" + name} ]; } else null
```
aanderse:
```suggestion
          if pkgs ? ${"xf86-video-" + name} then { modules = [ pkgs.${"xf86-video-" + name} ]; } else null
```

i don't think you want the `.` in the name

aanderse (changes requested):
> should I reopen the other PR or create a new one like in this case?

my preference would be a new one 👍

PR #64, opened 2026-05-10
modules/lxqt: add LXQt desktop environment

deathbymanatee:
Still kinda learning nix so feel free to suggest any changes if necessary. If anyone wants to try out the x11 session let me know, I haven't done much testing on it.

Module was tested on Wayland with `mdevd`, `seatd`, and the `labwc` compositor, and it works perfectly. I tested x11 with `openbox` with the same device / seat manager in virtualbox and it worked just as well if I recall correctly. 

Some of the package lists are a bit opinionated so I'm open to trimming it down if need be.

### DIFF
--- a/modules/programs/lxqt/default.nix
+++ b/modules/programs/lxqt/default.nix
@@ -1,0 +1,235 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  cfg = config.programs.lxqt;
+
+  inherit (pkgs) lxqt;
+  inherit (pkgs) kdePackages;
+
+  xSessionFile = pkgs.writeTextDir "share/xsessions/lxqt.desktop" ''
+    [Desktop Entry]
+    Name=LXQt (X11)
+    Comment=LXQt Desktop
+    Exec=${pkgs.lxqt.lxqt-session}/bin/startlxqt
+    Type=Application
+    DesktopNames=LXQt
+  '';
+
+  waylandSessionFile = pkgs.writeTextDir "share/wayland-sessions/lxqt-wayland.desktop" ''
+    [Desktop Entry]
+    Name=LXQt (Wayland)
+    Comment=LXQt Wayland Desktop
+    Exec=${pkgs.dbus}/bin/dbus-run-session -- ${pkgs.lxqt.lxqt-wayland-session}/bin/startlxqtwayland
+    Type=Application
+    DesktopNames=LXQt
+  '';
+
+  libinput = pkgs.libinput.override (
+    lib.optionalAttrs config.services.mdevd.enable {
+      udev = pkgs.libudev-zero;
+      wacomSupport = false;
+    }
+  );
+
+  # TODO: present in nixpkgs utils module, maybe port to finix?
+  removePackagesByName =
+    packages: packagesToRemove:
+    let
+      namesToRemove = map lib.getName packagesToRemove;
+    in
+    lib.filter (x: !(lib.elem (lib.getName x) namesToRemove)) packages;
+
+  packages = {
+    preRequisitePackages = [
+      kdePackages.kwindowsystem # provides some QT plugins needed by lxqt-panel
+      kdePackages.libkscreen # provides plugins for screen management software
+      pkgs.libfm
+      pkgs.libfm-extra
+      pkgs.menu-cache
+      kdePackages.qtsvg # provides QT plugins for svg icons
+    ];
+
+    corePackages = [
+      ### BASE
+      lxqt.libqtxdg
+      lxqt.libsysstat
+      lxqt.liblxqt
+      lxqt.qtxdg-tools
+      lxqt.libdbusmenu-lxqt
+
+      ### CORE 1
+      lxqt.libfm-qt
+      lxqt.lxqt-about
+      lxqt.lxqt-admin
+      lxqt.lxqt-config
+      lxqt.lxqt-menu-data
+      lxqt.lxqt-notificationd
+      lxqt.lxqt-openssh-askpass
+      lxqt.lxqt-powermanagement
+      lxqt.lxqt-qtplugin
+      lxqt.lxqt-sudo
+      lxqt.lxqt-themes
+      lxqt.pavucontrol-qt
+      lxqt.lxqt-session
+      lxqt.lxqt-wayland-session
+
+      ### CORE 2
+      lxqt.lxqt-panel
+      lxqt.lxqt-runner
+      lxqt.pcmanfm-qt
+
+      pkgs.xdg-utils
+      pkgs.libnotify
+    ]
+    ++ lib.optionals config.services.elogind.enable [
+      ### CORE 1
+      lxqt.lxqt-policykit
+    ]
+    ++ lib.optionals cfg.xsession.enable [
+      ### CORE 1
+      pkgs.lxqt.lxqt-globalkeys
+    ];
+
+    optionalPackages = [
+      ### LXQt project
+      lxqt.qterminal
+      lxqt.obconf-qt
+      lxqt.lximage-qt
+      lxqt.lxqt-archiver
+
+      ### QtDesktop project
+      lxqt.qps
+      lxqt.screengrab
+
+      ### Screen saver
+      pkgs.xscreensaver
+    ];
+  };
+
+in
+{
+  options.programs.lxqt = {
+
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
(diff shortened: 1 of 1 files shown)


modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      lxqt.liblxqt
+      lxqt.qtxdg-tools
+      lxqt.libdbusmenu-lxqt
+
+      ### CORE 1
+      lxqt.libfm-qt
+      lxqt.lxqt-about
+      lxqt.lxqt-admin
+      lxqt.lxqt-config
+      lxqt.lxqt-globalkeys
```
aanderse: i believe this is only applicable to x sessions, isn't it?

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      lxqt.lxqt-menu-data
+      lxqt.lxqt-notificationd
+      lxqt.lxqt-openssh-askpass
+      lxqt.lxqt-policykit
+      lxqt.lxqt-powermanagement
```
aanderse, line 72: does this work with `mdevd`?

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      lxqt.obconf-qt
+      lxqt.lximage-qt
+      lxqt.lxqt-archiver
+
+      ### QtDesktop project
+      lxqt.qps
+      lxqt.screengrab
+
+      ### Default icon theme
+      kdePackages.breeze-icons
```
aanderse: is this needed given we have an icon option?

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+        description = ''
+          The default Wayland compositor package to use.
+        '';
+      };
+    };
+
+    xsession = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
```
aanderse, line 171:
```suggestion
        default = config.services.xserver.enable or false;
```

maybe?

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages =
+      packages.preRequisitePackages
+      ++ packages.corePackages
+      ++ packages.optionalPackages
+      ++ [ cfg.iconThemePackage ]
+      ++ (removePackagesByName packages.optionalPackages cfg.excludePackages)
+      ++ cfg.extraPackages
+
+      # little messy, is there a better way to do this? lib.optional cfg.wayland.enable [...] results in a typeerror
```
aanderse: basically don't ever use `lib
optional` and use `lib.optionals` instead - they are very different

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      "/share"
+      "/share/icons"
+      "/share/pixmaps"
+    ];
+
+    security.pam.environment = {
+      XDG_CONFIG_DIRS.default = [ "/run/current-system/sw/share" ];
+    };
+
+    # probably not needed since default compositor is not kwin
```
aanderse: let's drop this comment as it isn't needed

on the future we could make adjustments to the option set to add support for kwin if desired

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+    #     capabilities = "cap_sys_nice+ep";
+    #     source = "${lib.getBin pkgs.kdePackages.kwin}/bin/kwin_wayland";
+    #   };
+    # };
+
+    xdg.portal.portals = [
+      pkgs.lxqt.xdg-desktop-portal-lxqt
+    ];
+
+    # this is also kind of messy
```
aanderse: maybe best to use `lib.optionalString` instead because it's entirely possible someone wants both x and `wayland`

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+
+    extraPackages = lib.mkOption {
+      type = with lib.types; listOf package;
+      default = [ ];
+      defaultText = lib.literalExpression "[ ]";
+      example = lib.literalExpression "with pkgs; [ lxqt.qterminal ]";
+      description = "Extra packages to be installed system wide.";
+    };
+
+    excludePackages = lib.mkOption {
```
aanderse, line 137: generally i dislike these `exclude*` type options because they imply the module is opinionated... but in this scenario i think the justification is valid: `lxqt` is a desktop environment... it is opinionated! so looks good 👍

aanderse (approved):
looks great, would merge as-is if desired

let me know if you want to make any changes first. thanks for the contribution!

modules/programs/lxqt/default.nix
deathbymanatee: I'm not entirely sure as I've only tested on my desktop, so I only get an error message saying that no power devices were found

modules/programs/lxqt/default.nix
deathbymanatee: GlobalKeys yes. I'll move that to the xsession dependent packages

deathbymanatee:
The x11 session currently doesn't start with just the .desktop entry so I'm gonna see if I can fix that before merging. Even though it doesn't work with `mdev` i would still like to be able to launch it so there isn't much extra configuration necessary on the user's part (e.g. an `.xinitrc` file)

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      pkgs.lxqt.xdg-desktop-portal-lxqt
+    ];
+
+    # this is also kind of messy
+    environment.etc."xdg/lxqt/session.conf".text =
+      if cfg.wayland.enable then
+        "COMPOSITOR=${cfg.wayland.compositor.pname}"
+      else
+        (if cfg.xsession.enable then "WINDOW_MANAGER=${cfg.xsession.windowManager.pname}" else "");
+  };
```
aanderse:
```suggestion
    environment.etc."xdg/lxqt/session.conf".source = (pkgs.formats.ini { }).generate "session.conf" {
      General =
        lib.optionalAttrs cfg.wayland.enable { compositor = cfg.wayland.compositor.pname; }
        // lib.optionalAttrs cfg.xsession.enable { window_manager = cfg.xsession.windowManager.pname; };
    };
```

looks like this information that i provided you was out of date - when i logged in i was told i need to select a compositor... i read the documentation and found this is what we should have

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+          [
+            (lib.hiPrio xSessionFile)
+            cfg.xsession.windowManager
+
+            # had issues without this package
+            pkgs.libxcb-cursor
+          ]
+        else
+          [ ]
+      );
```
aanderse:
```suggestion
      ++ (lib.optionals cfg.wayland.enable [
        (lib.hiPrio waylandSessionFile)
        cfg.wayland.compositor
      ])
      ++ (lib.optionals cfg.xsession.enable [
        (lib.hiPrio xSessionFile)
        cfg.xsession.windowManager

        # had issues without this package
        pkgs.libxcb-cursor
      ]);
```

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+      ### CORE 1
+      lxqt.libfm-qt
+      lxqt.lxqt-about
+      lxqt.lxqt-admin
+      lxqt.lxqt-config
+      lxqt.lxqt-globalkeys
+      lxqt.lxqt-menu-data
+      lxqt.lxqt-notificationd
+      lxqt.lxqt-openssh-askpass
+      lxqt.lxqt-policykit
```
aanderse: this needs to be conditional on `config.services.elogind.enable` - `polkit` agents can't run without a session manager

modules/programs/lxqt/default.nix
```diff
@@ -0,0 +1,239 @@
...
+    # probably not needed since default compositor is not kwin
+    # security.wrappers = {
+    #   kwin_wayland = {
+    #     owner = "root";
+    #     group = "root";
+    #     capabilities = "cap_sys_nice+ep";
+    #     source = "${lib.getBin pkgs.kdePackages.kwin}/bin/kwin_wayland";
+    #   };
+    # };
+
```
aanderse:
```suggestion
```

aanderse (changes requested):
i tested and have a few more notes

PR #66, opened 2026-05-12
modules/iwd: add `enableEad` as available option

deathbymanatee:
allows users to enable `iwd's` builtin ethernet authentication daemon


deathbymanatee:
I think my autoformat editor plugin also touched up the formatting, which is why there are a few more additional changes than just the additional option

deathbymanatee:
changing source branch

PR #71, opened 2026-05-22
fix(etc): /etc/os-release text updated

s0me1newithhand7s:
>[!TIP]
> roses are red,
> violets are blue,
> `/etc/os-release` is good,
> but there is always a room for improve. 

```Nix
ANSI_COLOR="0;38;2;231;56;71"
```
is true to `finix-logo` red color;
all links are valid for `/etc/os-release` strings;
not updating version itself;


aanderse:
thanks!

PR #72, opened 2026-05-22
modules/docker: init

deathbymanatee:
Adds docker as a service

### DIFF
--- a/modules/services/docker/default.nix
+++ b/modules/services/docker/default.nix
@@ -1,90 +1,135 @@
-# Docker service.
-
 {
   config,
-  lib,
   pkgs,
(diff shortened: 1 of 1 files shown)


deathbymanatee:
TODO: 
- [x] enableOnBoot option not implemented, but it starts on syslogd/ready by default
- [x] autoPrune option not implemented
- [x] implement autoprune service as cron job
- [ ] test autoprune
- ~~[x] add assertion that prevents any listenOptions with `fd://` as a prefix~~ users should be allowed to break their docker :D it's in the documentation that `fd://` is not supported so i will not bother enforcing it with an assertion

PR #73, opened 2026-05-24
modules/gardendevd: init

chocoblocko9:
havent tested this i just pulled stuff from my config out into a module lol, probably works

cc @aanderse and maybe @deathbymanatee for testing

### DIFF
--- a/modules/services/gardendevd/default.nix
+++ b/modules/services/gardendevd/default.nix
@@ -1,0 +1,199 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  cfg = config.services.gardendevd;
+
+  package = pkgs.gardendevd.overrideAttrs (old: {
+    version = "0.2-unstable-2026-07-03";
+
+    src = old.src.override {
+      tag = null;
+      rev = "ec73dc569382404bc6620c9857b7e09206bc282e";
+      hash = "sha256-8VOJFz5QtlyLbAf87rtNXSvnrfPoyQVAKwuD+YkfzdQ=";
+    };
+
+    mesonFlags = [
+      (lib.mesonEnable "dracut" false)
+      (lib.mesonEnable "uaccess" true)
+    ];
+  });
+in
+{
+  options.services.gardendevd = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable [gardendevd](${cfg.package.meta.homepage}) as a system service.
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = package;
+      defaultText = lib.literalExpression "pkgs.gardendevd";
+      description = ''
+        The package to use for `gardendevd`.
+      '';
+    };
+
+    debug = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable debug logging.
+      '';
+    };
+
+    extraArgs = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ ];
+      description = ''
+        Additional arguments to pass to `gardendevd`.
+      '';
+    };
+
+    path = lib.mkOption {
+      type = with lib.types; listOf path;
+      default = [ ];
+      description = ''
+        Packages added to the {env}`PATH` environment variable when
+        executing programs from udev rules.
+
+        coreutils, gnu{sed,grep}, util-linux and kmod are
+        automatically included.
+      '';
+    };
+  };
+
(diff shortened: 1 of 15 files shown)


modules/services/gardendevd/default.nix
```diff
@@ -0,0 +1,61 @@
...
+      description = ''
+        Log level
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    services.mdevd.nlgroups = lib.mkForce 2;
```
aanderse: ✔️

modules/services/gardendevd/default.nix
```diff
@@ -0,0 +1,61 @@
...
+
+    finit.services.gardendevd = {
+      description = "udev daemon running on top of mdevd to replace systemd-udev";
+      command = "${cfg.package}/bin/gardendevd -v " + lib.toString cfg.log-level;
```
aanderse:
```suggestion
      command = "${cfg.package}/bin/gardendevd -D %n -v " + lib.toString cfg.log-level;
```

modules/services/gardendevd/package.nix
```diff
@@ -0,0 +1,58 @@
...
+
+  src = fetchFromCodeberg {
+    owner = "Gardenhouse";
+    repo = "gardendevd";
+    rev = "v${finalAttrs.version}";
+    sha256 = "sha256-8G6Omeia1W+4dZOVHGtY/9CnKEpqD2x/W8Zkjt7fK/Q=";
+  };
+
+  nativeBuildInputs = [
+    acl
```
aanderse: belongs in `buildInputs`

modules/services/gardendevd/package.nix
```diff
@@ -0,0 +1,58 @@
...
+    elogind
```
aanderse: belongs in `buildInputs`, potentially guarded behind a top level argument

modules/services/gardendevd/package.nix
```diff
@@ -0,0 +1,58 @@
...
+    meson
+    ninja
+    pkg-config
+  ];
+
+  buildInputs = [ mdevd ];
```
aanderse: not needed here

modules/services/gardendevd/package.nix
```diff
@@ -0,0 +1,58 @@
...
+  elogind,
+  mdevd,
+  meson,
+  ninja,
+  pkg-config,
+  util-linux,
+  kmod,
+}:
+
+stdenv.mkDerivation (finalAttrs: {
```
aanderse: should add `mesonFlags`, force enable `mdevd` rules, etc...

modules/services/gardendevd/default.nix
aanderse: `notify = "s6";`

aanderse (commented):
just a few notes

deathbymanatee:
suggestion: udev has an attribute that allows for listing packages with udev rules that need to be dumped into /etc/udev/rules.d, i think there should be some level of equivalent functionality for this since gardendevd is trying to be a near drop-in replacement for (e)udev

aanderse:
> suggestion: udev has an attribute that allows for listing packages with udev rules that need to be dumped into /etc/udev/rules.d, i think there should be some level of equivalent functionality for this since gardendevd is trying to be a near drop-in replacement for (e)udev

`keventd` will also read `udev` rules - so what we really need is a device manager abstraction, under `providers.*` perhaps, where device managers can advertise whether they support `udev` vs `mdev` style rules... or both. then device managers can all share the same set of rules without needing to add multiple option sets.

i will need to give this some more thought.

N1meses:
Tested this pr on hardware: tmpfs root with by-partlabel mounts

Working: 
- boots to desktop -> by-partlablel mounts fine
- coldplug / hotplug
- inputs in niri and hwdb works
- Audio, NetworkManager 

One Bug found:
- /modules/programs/brightnessctl: gates udevSupport = config.services.udev.enable || config.services.keventd.enable, so gardendevd is missing so the package build without its udev rules

aanderse:
what is needed to merge this PR:

- fixup `brightnessctl` as per comments from @N1meses
- run `lon update nixpkgs`
- `package.nix` in favour of `pkgs.gardendevd`
- apply an `overrideAttrs` to bump versions to whatever is needed

@chocoblocko9 _if_ you have availability are you able to take these actions? :bowing_man:

aanderse:
merging from my system... running `gardendevd` :sunglasses:

thank you _sooo_ much for pushing this through to the finish line!

PR #76, opened 2026-05-25
feat(security): common-bwrap function init

s0me1newithhand7s:
could be used as package before executable in `finit.services.<...>`

### DIFF
--- a/modules/security/common-bwrap.nix
+++ b/modules/security/common-bwrap.nix
@@ -1,0 +1,58 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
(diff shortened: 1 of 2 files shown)


s0me1newithhand7s:
btw, i think it's better to create it as a lib, like `lib.bwrap.trap` to use it... wdyt?

aanderse:
currently not actionable - continue discussion and we can re-open if desired

s0me1newithhand7s:
recent pushes in https://github.com/s0me1newithhand7s/finix/tree/finix/bwrap, maybe re-open?

PR #77, opened 2026-05-25
modules/limine: remove unconditional reference to fwupd module

aanderse:
user reported an error today - i didn't get a chance to test out yet, but seems reasonable solution

@Ollie-Gutierrez - what do you think?


Ollie-Gutierrez (approved):
This looks good to me, no changes needed to the python script either, so definitely a better way to handle it 👍

PR #80, opened 2026-05-26
modules/iwd/enableEad: enable ethernet authentication daemon provided by iwd

deathbymanatee:
reopened from #66 to keep a clean main branch on my fork

### DIFF
--- a/modules/services/docker/default.nix
+++ b/modules/services/docker/default.nix
@@ -1,90 +1,135 @@
-# Docker service.
-
 {
   config,
-  lib,
   pkgs,
+  lib,
   ...
 }:
-
-with lib;
 
 let
 
   cfg = config.services.docker;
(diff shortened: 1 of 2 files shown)


aanderse:
i looked at the [upstream](https://git.kernel.org/pub/scm/network/wireless/iwd.git/tree/wired/ead.service.in) `systemd` service and noticed a few things we could add: 
- rlimits (nproc)
- capabilities

additionally i think this should be a separate service and likely requires minimal testing - do we have any ability to test at the moment?

deathbymanatee:
i could write something up, but i don't think i'll be able to verify functionality other than just "the service started and didn't crash"

aanderse:
without the ability to test this at all i suggest we close for now and can investigate when someone has capacity or capability to test

Issue #79, opened 2026-05-26
[RFC] coordinating efforts to write, maintain, and centralize `finix` documentation


deathbymanatee:
NixOS has a bit of a documentation problem. while the official manual is well written, it can sometimes feel far to technical for the average user to digest. this was certainly my experience; once i followed the initial installation guide, i hardly ever used the official manual aside for some miscellaneous configuration options not covered in the wikis. 

`finix` has a real opportunity to learn from the mistakes of NixOS and have one centralized place where all of its documentation lives -- user guides, technical manuals, and the like -- instead of having two separate wikis that may or may not be up to date at any given time, or a technical manual that is entirely separate from both wikis. so, here are my current suggestions: 

in the early stages (i.e. right now), we can utilize github's Wiki feature to write some initial documentation covering essential topics, such as: 
- installation instructions (already covered by Vitrial over on codeberg, but would be great if it was migrated over to the main repository)
- explanation of profiles and how to set them up
- setting up a minimal system without profiles
- FAQ
    - aside: a common question in the reddit thread posted by @ fire flip was: "can I port my existing nixos configuration over? what is the level of compatibility between the two?" i.e. what are the main differences between the two.

as the userbase grows, we can start exploring alternative documentation hosting solutions that are easier for search engines to index to enhance discoverability. while the discord is great for individual support, i generally think discord is where documentation goes to die; it is not accessible outside of discord, not intuitive to use, and not everyone is willing to join a public server. 

let me know any thoughts. i am willing to write and maintain some of these documents to help get this off the ground. it's probably a little early to bring up this sort of issue given `finix` doesn't have an established release model just yet, but we already have plenty of users (myself inlcuded) daily driving this distro so the sooner we get some docs going, the better. 

i think the options wiki can stay how it is, but it should be linked somewhere in the repo. @aanderse already addressed this in the discord but i figured it would be appropriate to repeat here

PR #82, opened 2026-05-27
Add install guides and documentation links into README

deathbymanatee:
#79


aanderse:
this is awesome!

there are a few things we need to fixup on the example configs, though - like `programs.limine.force = true;`... we really need to get that out of there before we merge this :sweat_smile:

PR #83, opened 2026-05-27
modules/pipewire: init

deathbymanatee:
set up as a program with all the extra configuration options present in nixos. builds and runs. i have no idea what to do about the config options using pkgs.buildEnv or pipewire upstream setting up environment variables. wireplumber upstream also wants to add onto XDG_DATA_DIRS as that is where wireplumber looks for extra configuration scripts. 

opening as draft for now until we get more consensus on if to even include this as a core program module in the first place.


deathbymanatee:
should be ready for review now

aanderse:
`pipewire` and `wireplumber` modules have been added

PR #85, opened 2026-05-30
modules/limine: fixes

chocoblocko9:
1. Fixes the documentation links being dead for the `limine` module
2. More importantly, fixes the wallpapers not being entered into the config right. an honest mistake :)


aanderse (approved):
i just ran into this yesterday as well... thanks!

PR #88, opened 2026-06-01
inital work for finixModules

parzivale:
The idea behind this draft is to more clearly label finixModules as seperate from nixosModules to attempt to stop users to cross contaminating modules where possible, though any upstream nixos modules that does not set meta.class = "nixos" will still function with these changes

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
 let
   flake = import ./flake.nix;
 
-  self = flake.outputs { inherit self; };
+  self = flake.outputs {inherit self;};
 in
-self
+  self
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
 let
   flake = import ./flake.nix;
 
-  self = flake.outputs {inherit self;};
+  self = flake.outputs { inherit self; };
 in
-  self
+self
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,7 @@
   outputs =
     { self }:
     {
-      nixosModules = import ./modules;
+      finixModules = import ./modules;
 
       lib.finixSystem =
         {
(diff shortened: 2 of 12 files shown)


parzivale:
Ah it looks like I was mistaken its actually _class not meta.class but the other points still stand

parzivale:
Potentially we could set up a seperate evalModules call with the class set to nixos to allow users to import nixos modules into their finix configuration

this would be a little hacky to setup but it would be the most reliable way to ensure finix users can import nixos modules at their discretion without allowing nixos users to import finix modules

something like 


```nix
nixosModules = lib.evalModules {class = "nixos"; modules = [ <All collected nixos module imports>]};
```

then we could extract the evaluated config options to use in finix

parzivale:
ok using this latest commit you should be able to register nixos modules for finix like so

```nix
 # modules/services/nginx.nix (a finix module)
  { pkgs, ... }: {
    nixosOptions = [
        "${pkgs.path}/nixos/modules/services/web-servers/nginx.nix"
    ];

    nixosModules = {
      services.nginx.enable = true;
      services.nginx.virtualHosts."example.com" = {
        root = "/var/www";
      };
    };
  }
```

parzivale:
ok slight changes now it looks like this

Module definition (a finix module that wraps a known-good NixOS module):
```nix
  # modules/services/nginx.nix
  { pkgs, ... }: {
    nixos.options = [
      "${pkgs.path}/nixos/modules/services/web-servers/nginx.nix"
    ];
  }
```
User implementation (a finix system using that module):
```nix

  # configuration.nix
  { pkgs... }: {
    imports = [ ./modules/services/nginx.nix ];

    nixos.config = {
      services.nginx.enable = true;
      services.nginx.virtualHosts."example.com" = {
        root = "/var/www";
      };
    };
  }
  ```

parzivale:
Im open to changing the naming as im not sure if what ive done makes sense but i think this is good for now

parzivale:
stalled until https://github.com/NixOS/nixpkgs/issues/530003 is addressed

PR #89, opened 2026-06-02
modules/nix-channel: init

aanderse:


modules/programs/nix-channel/default.nix
```diff
@@ -0,0 +1,48 @@
...
+      ''f /root/.nix-channels - - - - ${lib.concatStringsSep "\\n" channels}\n''
+    ];
+
+    security.pam.environment = lib.optionalAttrs cfg.enable {
+      NIX_PATH.default = [
+        # channel aliases
+        "nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos"
+
+        # instruct nixos-rebuild where configuration is located
+        "nixos-config=/etc/nixos/configuration.nix"
```
aanderse, line 40: actually... seems likely this should be `/etc/nixos/system.nix`, or maybe even `/etc/finix/system.nix` - we can also pre-populate that file with a `tmpfiles` rule!

modules/services/nix-daemon/default.nix
```diff
@@ -65,6 +66,8 @@ let
     '';
 in
 {
+  imports = [ modules.nix-channel ];
```
aanderse, line 69: future work: create a `programs.nix` module and move this import to that module

JumpIn-Git:
hey, it would be a great idea to make a NIX_PATH option here, which would help pr #195

modules/programs/nix-channel/default.nix
JumpIn-Git: i would reccomend moving it to a option which is null by default

aanderse: how do you envision that looking like?

JumpIn-Git: > how do you envision that looking like?

something like programs.nix.nixos-config = "/etc/finix/config.nix"

PR #96, opened 2026-06-08
programs/doas: added common options

xZecora:


aanderse:
aren't these options per rule in `doas`?

xZecora:
this adds them to the existing rule for the wheel group, should I clarify this in the options descriptions?

modules/programs/doas/default.nix
```diff
@@ -6,6 +6,10 @@
 }:
 let
   cfg = config.programs.doas;
+
+  optionalStringElse =
```
aanderse: i would prefer to avoid indirection like this

modules/programs/doas/default.nix
```diff
@@ -29,6 +33,30 @@ in
         The package to use for `doas`.
       '';
     };
+
+    requirePassword = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        				Whether or not to prompt the user for their password.
```
aanderse:
```suggestion
        Whether or not to prompt the user for their password.
```

`spaces`, not `tabs` please

modules/programs/doas/default.nix
```diff
@@ -29,6 +33,30 @@ in
...
+        			'';
+    };
+
+    persist = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        				Whether or not to allow credentials to persist for users for 5 minutes.
```
aanderse:
```suggestion
        Whether or not to allow credentials to persist for users for 5 minutes.
```

`spaces`, not `tabs` please

modules/programs/doas/default.nix
```diff
@@ -29,6 +33,30 @@ in
...
+        			'';
+    };
+
+    keepEnv = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        				Whether or not to keep the users environment during privilege escalation.
```
aanderse:
```suggestion
        Whether or not to keep the users environment during privilege escalation.
```

`spaces`, not `tabs` please

aanderse (changes requested):
> should I clarify this in the options descriptions?

yes, please

PR #103, opened 2026-06-17
hardware/microcode: added updateMicrocode options for AMD and Intel CPUs

xZecora:


aanderse:
i will test this (on `intel`) at some point... sorry for the delay :bowing_man:

PR #116, opened 2026-06-28
activation/etc: rewrote etc-setup script to be POSIX shell

xZecora:


modules/environment/etc/setup-etc.sh
```diff
@@ -0,0 +1,123 @@
...
+  fi
+done
+
+# Delete files that were copied in a previous version but not in the current.
+while [ "$1" != "" ]; do
+  if [ -n "$1" ]; then
+    shift
+    continue
+  fi
+
```
aanderse: doesn't the `while` imply the `if` will always evaluate `true`? _if_ so, that means everything after here is dead code

you can test this by adding `environment.etc.foo.text = "bar";`, switching to a generation with this, and then removing and switching again

modules/environment/etc/setup-etc.sh
```diff
@@ -0,0 +1,123 @@
...
+etcDirs="/tmp/etcDirs"
+
+mkfifo $etcDirs
+
+find "$etc" -mindepth 1 >$etcDirs &
+
+# For every file in the etc tree, create a corresponding symlink in /etc.
+# Use process substitution to avoid subshell issues with pipes.
+cat $etcDirs | while IFS= read -r entry; do
+  fn="$(echo "$entry" | @sed@ "s,${etc},,")"
```
aanderse: isn't this `posix`? `fn="${entry#"$etc"/}"`

modules/environment/etc/default.nix
```diff
@@ -198,7 +199,7 @@ in
     # TODO: create an alternative implementation with... https://github.com/Gerg-L/linker
     system.activation.scripts.etc = lib.stringAfter [ "users" ] ''
       echo "setting up /etc..."
-      ${pkgs.perl.withPackages (p: [ p.FileSlurp ])}/bin/perl ${./setup-etc.pl} ${buildEtc}/etc
+      ${pkgs.dash}/bin/dash ${setup-etc} ${buildEtc}/etc
```
aanderse: already running under `bash` so we don't specifically need to call an interpreter **if** `${setup-etc}` is executable... 🤔

modules/environment/etc/setup-etc.sh
```diff
@@ -0,0 +1,123 @@
...
+  while IFS= read -r line; do
+    set -- "$@" "$line"
+  done </etc/.clean
+fi
+
+clean_tmp="$(mktemp)"
+created_tmp="$(mktemp)"
+etcDirs="/tmp/etcDirs"
+
+mkfifo $etcDirs
```
aanderse: is this actually needed?

modules/environment/etc/setup-etc.sh
xZecora: You're entirely correct. Just corrected this in latest commit. For future reference though, I found that `environment.etc.foo.text = "bar";` creates a symlink and not a file. So it doesn't actually test this part of the function script.

modules/environment/etc/setup-etc.sh
xZecora: Yes... I think I psyched myself out on this one. I would have sworn I tested it... Either way, there goes the sed dependency! Thanks.

modules/environment/etc/setup-etc.sh
xZecora: Yes and no, if `/etc` is large enough (like scary big) `find` can become a bottleneck, this lets you start the loop before `find` finishes executing. I switched in the current version to just be a normal output redirect and input redirect because I do think that this is not technically safe and kinda gross looking

modules/environment/etc/default.nix
xZecora: Currently it isn't executable. Is there a safe way to set that?

N1meses:
I tested this pr and it holds its promise.
one issue which @aanderse mentioned in a comment before:
in 'setup-etc.sh' l55:
```sh
fn="${entry#"${etc}"}"          # → "/doas.conf"   (leading slash)
```
it should be:
```sh
fn="${entry#"${etc}"/}"
```
if the slash is missing it creates the files and subsequently removes them again because:
```sh
if ! grep -qxF "$1" "$created_tmp"; then    # "doas.conf" vs "/doas.conf" → never matches
  rm -f "/etc/$1"
fi
```

xZecora:
I corrected @N1meses comment, thanks for catching that. Not sure why I never had issues with it.

aanderse:
i would love to finally drop our `perl` dependency... have we tested this with `busybox` for `coreutils`?

@N1meses have you tested this under a variety of situations? we're good to merge?

PR #117, opened 2026-06-29
programs/shadow: added pam_limits call to login pam service

xZecora:


aanderse:
additionally we [could/should](https://wiki.debian.org/Limits) add to `su`, `sudo`, `cron`, `fcron`, `atd`, `sshd`, `greetd`, `ly`, and `lemurs`...

xZecora:
skipped `services/sddm` because the structure was different and i'm not sure the correct way to implement it there. if someone else could take a look that would be great

aanderse (approved):
amazing! thank you _so_ much for this! 🚀

PR #119, opened 2026-07-03
modules/virtualisation/qemu: fall back when   hostPlatform.linux-kernel is absent

N1meses:
Problem:
nixpkgs removed linux-kernel from platform elaboration during the 26.11 cycle (https://github.com/NixOS/nixpkgs/pull/530133). On any nixos-unstable from roughly mid-June onward, evaluating a VM built via tests/lib fails:
```
error:
       … while checking flake output 'packages'

       … while checking the derivation 'packages.x86_64-linux.vm-icarus'

       … while calling the 'derivationStrict' builtin
         at «nix-internal»/derivation-internal.nix:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating the option `virtualisation.qemu.argv':

       … while evaluating definitions from `/nix/store/5dbfplbzn8p42s9q1rsrrvjqn9pndi7i-source/modules/virtualisation/qemu.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute 'linux-kernel' missing
       at «github:finix-community/finix/6f527e0439a16d03e4f3889b1c2a6b35404d1004?narHash=sha256-RXvEgkVX69kQzT7sgF40VuALqMVk5degJzgWG/nOC9w%3D»/modules/virtualisation/qemu.nix:112:47:
          111|       "-kernel"
          112|       "${config.boot.kernelPackages.kernel}/${pkgs.stdenv.hostPlatform.linux-kernel.target}"
             |                                               ^
          113|       "-initrd"
```
The Modules modules/boot/bootspec.nix and modules/system/activation/default.nix handle this already with the fallback:

config.boot.kernelPackages.kernel.target or pkgs.stdenv.hostPlatform.linux-kernel.target

qemu.nix:112 was the one remaining call site without it.

### DIFF
--- a/modules/virtualisation/qemu.nix
+++ b/modules/virtualisation/qemu.nix
@@ -109,7 +109,7 @@ let
   bootModeArgs = {
     kernel = [
       "-kernel"
-      "${config.boot.kernelPackages.kernel}/${pkgs.stdenv.hostPlatform.linux-kernel.target}"
+      "${config.boot.kernelPackages.kernel}/${config.boot.kernelPackages.kernel.target or pkgs.stdenv.hostPlatform.linux-kernel.target}"
       "-initrd"
       "${config.boot.initrd.package}/initrd"
       "-append"
--- a/modules/virtualisation/qemu.nix
+++ b/modules/virtualisation/qemu.nix
@@ -109,7 +109,9 @@ let
   bootModeArgs = {
     kernel = [
       "-kernel"
-      "${config.boot.kernelPackages.kernel}/${config.boot.kernelPackages.kernel.target or pkgs.stdenv.hostPlatform.linux-kernel.target}"
+      "${config.boot.kernelPackages.kernel}/${
+        config.boot.kernelPackages.kernel.target or config.boot.kernelPackages.stdenv.hostPlatform.linux-kernel.target
+      }"
       "-initrd"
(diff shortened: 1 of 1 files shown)


aanderse:
thanks for contributing a fix! it looks like i missed this back in https://github.com/finix-community/finix/commit/8aa84374e429d95e2d760da1b6bb8c829c7adb53 :sweat_smile:

should we be using `config.boot.kernelPackages.stdenv.hostPlatform.linux-kernel.target` or `pkgs.stdenv.hostPlatform.linux-kernel.target`? i imagine they are always the same... but it would be nice to be consistent

thoughts?

N1meses:
They're identical whenever kernelPackages comes from the evaluating nixpkgs. They differ when boot.kernelPackages is overridden with a kernel set from another nixpkgs instance. Since target is a property of the kernel build, the kernel set's own stdenv.hostPlatform is the right source. It's also the only variant that still evaluates in the case the fallback exists for an older kernel set on a current nixpkgs. The fallback would throw exactly when it's needed. So: config.boot.kernelPackages.stdenv.hostPlatform.linux-kernel.target, matching bootspec.nix.

aanderse:
thank you!

PR #121, opened 2026-07-04
Modules/firewall

parzivale:
First attempt at a firewall module, needs more testing


modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,151 @@
...
+  };
+in
+{
+  imports = [
+    ./iptables.nix
+    ./nftables.nix
+  ];
+
+  options.providers.firewall = {
+    enable = lib.mkOption {
```
aanderse: please take a look at how other providers are defined - entirely separate from any implementation

modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,151 @@
...
+      default = false;
+      description = ''
+        Whether to enable the firewall. This is a simple stateful firewall that
+        blocks connection attempts to unauthorised TCP or UDP ports on this machine.
+      '';
+    };
+
+    backend = lib.mkOption {
+      type = lib.types.enum [
+        "iptables"
```
aanderse: should be left to individual implementations

modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,151 @@
...
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = if cfg.backend == "nftables" then pkgs.nftables else pkgs.iptables;
+      defaultText = lib.literalExpression "pkgs.nftables or pkgs.iptables depending on backend";
+      description = ''
+        The package to use for running the firewall service.
+      '';
+    };
+
+    trustedInterfaces = lib.mkOption {
```
aanderse: providers should be narrow in scope generally... but a firewall likely needs a bit more than the average provider - what do you think the full scope should be for this provider

modules/providers/firewall/iptables.nix
```diff
@@ -0,0 +1,92 @@
...
+    ${iptablesRestore}  /etc/iptables/rules.v4
+    ${ip6tablesRestore} /etc/iptables/rules.v6
+  '';
+in
+{
+  config = lib.mkIf (cfg.enable && cfg.backend == "iptables") {
+    environment.etc."iptables/rules.v4".text = rulesV4;
+    environment.etc."iptables/rules.v6".text = rulesV6;
+
+    finit.tasks.firewall = {
```
aanderse: this won't work - you need more run levels than `S` or three firewall won't be applied to an already running system

you want a remain after exit type task (`remain`) with a `post` script

modules/providers/firewall/nftables.nix
```diff
@@ -0,0 +1,67 @@
+{
```
aanderse: note there is already an `nftables` module so we should utilize that... or replace it

tests/firewall.nix
```diff
@@ -0,0 +1,101 @@
+# test for the providers.firewall module
```
aanderse, line 1: 🤩

thank you!

aanderse (changes requested):
please keep commits on topic, no need for `nix-daemon` path commits in a firewall PR

parzivale:
Yep I realised I had an old commit in my tree, I reverted it

modules/providers/firewall/default.nix
parzivale: I see the other implementations don't have an enable option, but a firewall is kind of also a service? How would you want people to turn off the firewall? Maybe an option under networking.firewall?

modules/providers/firewall/default.nix
parzivale: Wdym should be left to individual implementations?

modules/providers/firewall/default.nix
parzivale: Hmm I agree with keeping the scope minimal, I was just copying nixpkgs here as they trust the loopback interface unconditionally. I think it's a reasonable thing to keep though as allowing interfaces past the firewall feels like something we should keep here

modules/providers/firewall/iptables.nix
parzivale: Thanks! I'll update it

modules/providers/firewall/nftables.nix
parzivale: Gotcha I'll use it as that feels more correct, is it ok to add a dependency on nftables here then?

modules/providers/firewall/default.nix
aanderse: see implementations of `scheduler` provider like `cron` and `fcron` - very standard file and option layout you can follow

the provider isn't the service, it's a `trait`... the specific implementation is the `impl`

modules/providers/firewall/default.nix
aanderse: see `fcron` and `cron` folders, as well as `scheduler` provider (see `backend` only has `none` value there)

aanderse: see `fcron` and `cron` folders, as well as `scheduler` provider (see `backend` only has `none` value there)

modules/providers/firewall/default.nix
aanderse: specific implementations of firewall would have more options than the provider itself... i don't do a great job of demonstrating that in all the providers yet but what i mean is that the `nftables` and `iptables` services should have more options than the `firewall` trait

modules/providers/firewall/nftables.nix
aanderse: the functionality should be merged into the existing `nftables` module

modules/providers/firewall/default.nix
parzivale: Gotcha! Thanks for explaining in rust terms lmao. Where would you want the impl to live?

parzivale: Wait I think i get it!

parzivale:
@aanderse I believe I have addressed most of you concerns in this most recent commit, I also took the opportunity to make an iptables module and expand the nftables module

modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,106 @@
...
+  options.providers.firewall = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable the firewall. This is a simple stateful firewall that
+        blocks connection attempts to unauthorised TCP or UDP ports on this machine.
+      '';
+    };
+
```
aanderse, line 21:
```suggestion
```

providers should never be explicitly enabled or disabled, the `backend` option in combination with a `...$backend.enable = true|false` is sufficient

modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,106 @@
...
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        If set, refused packets are rejected rather than dropped. This sends an
+        ICMP "port unreachable" error back to the client (or a TCP RST for
+        existing connections), allowing clients to fail fast rather than timeout.
+        Rejecting packets makes port scanning somewhat easier.
+      '';
+    };
+
```
aanderse, line 41:
```suggestion
```

as mentioned over discord this should likely be a module specific option as i can't see (m)any system level modules needing to tweak this... or even being able to tweak this!

modules/providers/firewall/default.nix
```diff
@@ -0,0 +1,106 @@
...
+      description = ''
+        Range of open UDP ports.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = cfg.backend != "none";
```
aanderse, line 98: it is possible the user does not want a firewall, so we can turn this into a warning **if** rules are actually specified

modules/services/nftables/default.nix
```diff
@@ -26,22 +79,121 @@ in
       '';
     };
 
+    checkRuleset = lib.mkOption {
```
aanderse, line 82: i'll be honest... i am not a fan of these types of options. they cannot be done cleanly and are a best effort... i would suggest you drop

modules/services/nftables/default.nix
```diff
@@ -26,22 +79,121 @@ in
...
+      description = ''
+        Run `nft check` on the ruleset to spot syntax errors during build.
+        Because this is executed in a sandbox, the check might fail if it requires
+        access to any environmental factors or paths outside the Nix store.
+        To circumvent this, the ruleset file can be edited using the preCheckRuleset
+        option to work in the sandbox environment.
+      '';
+    };
+
+    checkRulesetRedirects = lib.mkOption {
```
aanderse, line 94: same comment as above, i suggest you drop

modules/services/nftables/default.nix
```diff
@@ -26,22 +79,121 @@ in
...
+          "/etc/services" = config.environment.etc.services.source;
+        }
+      '';
+      description = ''
+        Set of paths that should be intercepted and rewritten while checking the ruleset
+        using `pkgs.buildPackages.libredirect`.
+      '';
+    };
+
+    preCheckRuleset = lib.mkOption {
```
aanderse, line 116: same comment as above, i suggest you drop

modules/services/nftables/default.nix
```diff
@@ -26,22 +79,121 @@ in
...
+        }
+      '';
+      description = ''
+        Tables to be added to the ruleset.
+        Tables will be added together with delete statements to clean up the
+        table before every update.
+      '';
+    };
+
     configFile = lib.mkOption {
```
aanderse, line 167: i think we should _eventually_ drop this option entirely from this module... it entirely goes against the point of it. let's leave it for now but add a warning and then remove in a month or two.

modules/services/nftables/default.nix
```diff
@@ -26,22 +79,121 @@ in
...
+          ${deletions}
+          ${lib.concatStringsSep "\n" (
+            lib.mapAttrsToList (_: table: ''
+              table ${table.family} ${table.name} {
+                ${table.content}
+              }
+            '') enabledTables
+          )}
+        '';
+        checkPhase = lib.optionalString cfg.checkRuleset ''
```
aanderse, line 181: we should drop this, see reasoning above

modules/services/nftables/default.nix
```diff
@@ -52,10 +204,11 @@ in
     environment.systemPackages = [ cfg.package ];
 
     finit.tasks.nftables = {
+      runlevels = "23";
```
aanderse, line 207: hmm... i think we need this in more than just `2` or `3`... maybe omit `runlevels` entirely and use the default `234`?

aanderse (changes requested):
sorry for taking so long...

PR #124, opened 2026-07-07
fix(initrd): resolve blockdev race conditions and fix mdevd shell compatibility

FixeQD:
This PR bundles several fixes addressing race conditions during stage-1 block device discovery and mounting, alongside an initrd shell environment fix.

Changes breakdown:
1. **Robust Stage-1 Mounting:** Added `set -e` to `finix-mount-all` to ensure immediate failure propagation. Implemented a retry loop (30 attempts with 1s sleep) for the mounting sequence because a successful `mdevd-coldplug` only guarantees that uevents were triggered, not that the kernel has fully exposed the device nodes yet. The individual mount blocks now utilize `mountpoint -q` to prevent duplicate mounts and include a small timeout loop checking for device node existence
2. **Mdevd Script Compatibility:** Swapped out `pkgs.writeShellScript` for `pkgs.writeScript` with an explicit `#!/bin/sh` shebang. `writeShellScript` bakes in a hardcoded `/nix/store/...-bash` interpreter, which is absent in the initrd environment. 
3. **Enhanced Disk Symlinks:** Added a short retry loop for `blkid` inside `mdevd-disk.sh` to prevent empty lookups when the uevent fires before the partition table is fully parsed. Also introduced proper creation and cleanup for `by-partlabel` and `by-partuuid` symlinks.
4. **Switch-root Recovery:** Appended the `notty` flag to the rescue TTY declaration to prevent device-manager allocation issues if the initrd switch-root fails.

### DIFF
--- a/modules/finit/initrd.nix
+++ b/modules/finit/initrd.nix
@@ -17,11 +17,14 @@ let
     + (
       if builtins.elem "bind" mnt.options then
         ''
-          mount -o ${lib.concatStringsSep "," mnt.options} "$targetRoot${mnt.device}" "$targetRoot${mnt.mountPoint}"
+          mountpoint -q "$targetRoot${mnt.mountPoint}" || mount -o ${lib.concatStringsSep "," mnt.options} "$targetRoot${mnt.device}" "$targetRoot${mnt.mountPoint}"
         ''
       else
         ''
-          mount -t ${mnt.fsType} -o ${lib.concatStringsSep "," mnt.options} "${mnt.device}" "$targetRoot${mnt.mountPoint}"
+          mountpoint -q "$targetRoot${mnt.mountPoint}" || {
+            [ -e "${mnt.device}" ] # let finix-mount-all's outer retry handle waiting
+            mount -t ${mnt.fsType} -o ${lib.concatStringsSep "," mnt.options} "${mnt.device}" "$targetRoot${mnt.mountPoint}"
+          }
         ''
     );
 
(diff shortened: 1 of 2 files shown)


modules/finit/initrd.nix
```diff
@@ -268,7 +268,7 @@ in
           ''}
 
           run [1] name:switch-root finix-switch-root
-          tty [1] <run/switch-root/failure> @console rescue
```
aanderse, line 269: i'm not sure about using `notty` here - is `rescue` not sufficient?

FixeQD: Partially fixed my keyboard issue 
But mainly the issue was that the mdevd was missing so I can remove `notty` from here

FixeQD:
Some things got messy (cuz I mixed up branches) and I fixed it :c

modules/services/mdevd/default.nix
```diff
@@ -47,37 +47,56 @@ let
...
-              ;;
-          esac
-        done
+        # mdevd blocks its whole event loop until this script exits, so the blkid retry wait must run in the background, not foreground.
+        (
+          info=""
+          for _try in 1 2 3 4 5; do
+            info=$(blkid --output export "/dev/$MDEV" 2>/dev/null)
+            [ -n "$info" ] && break
+            sleep 0.2
```
aanderse, line 71: TODO: confirm if `busybox` is built with fractional second `sleep` support

modules/services/mdevd/default.nix
```diff
@@ -47,37 +47,56 @@ let
...
+          done
+
+          echo "$info" | while IFS='=' read -r key value; do
```
aanderse, line 74:
```suggestion
          print '%s\n' "$info" | while IFS='=' read -r key value; do
```

aanderse (commented):
i squashed commits as there was some back and forth, and removed profanity... hope you don't mind 🙇‍♂️

left 2 small notes we could/should probably follow up on at some point, but this is great! thanks again! 🚀

PR #125, opened 2026-07-07
feat(filesystems/swap): add discardPolicy, finit-managed randomEncryption, and boot.resumeDevice

FixeQD:
Ported standard NixOS swap and hibernation options to work seamlessly within the finix architecture, replacing systemd components with native finit configuration

Changes:
1. **Swap Discard Policy:** Backported the `swapDevices.*.discardPolicy` option (`once`, `pages`, `both`), mapping it correctly to standard swap mount options
2. **Finit-Managed Random Encryption:** Reimplemented NixOS-style random swap encryption without relying on systemd-cryptsetup targets. Instead of using static `/etc/fstab` lines, encrypted swaps are now generated imperatively via dedicated stage-1 `finit.tasks`. The tasks execute a lightweight script shell sequence on every boot.Added an assertion to ensure these devices use stable persistent paths rather than volatile labels or UUIDs
3. **Hibernation Support:** Added the `boot.resumeDevice` option to inject the `resume=` string into `boot.kernelParams`


modules/filesystems/options.nix
```diff
@@ -143,6 +143,84 @@ let
...
+          default = { };
+          description = ''
+            Encrypt swap device with a random key. This way you won't have a
+            persistent swap device, the entire contents become unreadable
+            after every reboot. Mutually exclusive with {option}`label`,
+            since the underlying mapped device is regenerated on every boot
+            and is therefore not activated through `/etc/fstab` like a
+            regular swap device, but through a dedicated `finit` task.
+          '';
+          type = lib.types.submodule {
```
aanderse: no specific need for a `submodule` here, right? let's save the evaluation and replace `randomEncryption = libmkOption {` with `randomEncryption = {`, yeah?

modules/filesystems/default.nix
```diff
@@ -98,6 +102,44 @@ let
...
+            set -eu
+            ${pkgs.cryptsetup}/bin/cryptsetup plainOpen \
+              -c ${lib.escapeShellArg re.cipher} \
+              -s ${toString re.keySize} \
+              ${lib.optionalString (re.sectorSize != 0) "--sector-size ${toString re.sectorSize}"} \
+              ${lib.optionalString re.allowDiscards "--allow-discards"} \
+              -d ${lib.escapeShellArg re.source} \
+              ${lib.escapeShellArg sw.device} ${lib.escapeShellArg name}
+            ${pkgs.util-linux}/bin/mkswap /dev/mapper/${name}
+            ${pkgs.util-linux}/bin/swapon -o ${lib.escapeShellArg (lib.concatStringsSep "," options)} /dev/mapper/${name}
```
aanderse: will `util-linuxMinimal` be sufficient here?

modules/filesystems/default.nix
```diff
@@ -98,6 +102,44 @@ let
...
+        options = sw.options
+          ++ lib.optional (sw.priority != null) "pri=${toString sw.priority}"
+          ++ lib.optional (sw.discardPolicy != null) (
+            if sw.discardPolicy == "both" then "discard" else "discard=${sw.discardPolicy}"
+          );
+      in
+      {
+        description = "Encrypted swap device on ${sw.device}";
+        runlevels = "S";
+        command = toString (
```
aanderse: i wonder if it is reasonable to avoid wrapping this in a shell script and simply inlining the command?

modules/filesystems/default.nix
```diff
@@ -98,6 +102,44 @@ let
...
+
+  # Swap entries with randomEncryption.enable can't be stable fstab lines: the backing device is a fresh /dev/mapper/<name> created with a brand new random key on every boot, so they're set up imperatively instead
+  isEncryptedSwap = sw: sw.randomEncryption.enable;
+  plainSwapDevices = lib.filter (sw: !isEncryptedSwap sw) config.swapDevices;
+  encryptedSwapDevices = lib.filter isEncryptedSwap config.swapDevices;
+
+  sanitizeName = s: lib.replaceStrings [ "/" " " ] [ "-" "-" ] (lib.removePrefix "/" s);
+
+  makeEncryptedSwapTask = sw: rec {
+    name = "cryptswap-${sanitizeName sw.device}";
```
aanderse:
```suggestion
  makeEncryptedSwapTask = sw: let name = "cryptswap-${sanitizeName sw.device}"; in {
    inherit name;
```

i think in this context `rec` makes it harder to reason about the code... so i would _suggest_ this change, though feel free to disagree and ignore this comment as it is more _opinion_ than actual code _quality_ review 🙇‍♂️

aanderse (commented):
i will admit i don't run encrypted swaps or hibernate my laptop - it has never been a need i've had... that being said... this all _looks_ quite reasonable

i left some very minor comments i would appreciate if you could look over

modules/filesystems/default.nix
FixeQD: Hmmm... Inlining means chaining 3 commands with their own quoting inside one `bash -c` string, which gets messy fast so I would keep the `writeShellScript` wrapper here

modules/filesystems/default.nix
```diff
@@ -96,6 +102,50 @@ let
       )
       + "\n"
     ) fstabFileSystems;
+
+  # Swap entries with randomEncryption.enable can't be stable fstab lines: the backing device is a fresh /dev/mapper/<name> created with a brand new random key on every boot, so they're set up imperatively instead
+  isEncryptedSwap = sw: sw.randomEncryption.enable;
+  plainSwapDevices = lib.filter (sw: !isEncryptedSwap sw) config.swapDevices;
+  encryptedSwapDevices = lib.filter isEncryptedSwap config.swapDevices;
+
+  sanitizeName = s: lib.replaceStrings [ "/" " " ] [ "-" "-" ] (lib.removePrefix "/" s);
```
aanderse, line 111: @N1meses this can be an issue, right?

N1meses: Here in some cases yes ( "//", "/ ", "  ",  " /"), but if only one of  " "  "/"  is present then its fine.

Also this wouldn't cause boot hang like in the other PR, because nothing is referencing this tasks success.

on the other hand sanitizeName has another issue: 
/dev/a-b → dev-a-b
/dev/a/b → dev-a-b
mount.nix carries uniqueness assertion, the swap doesn't

PR #126, opened 2026-07-07
feat(boot): add external bootloader install hook abstraction

FixeQD:
Ported the external bootloader hook pattern to finix and cleaned up duplicate options. This adds the `boot.loader.external` module to route custom lifecycle installation scripts directly to `providers.bootloader.installHook` while automatically managing the `efivarfs` mount, allowing the redundant EFI option declarations to be safely removed from the Limine module.


aanderse (commented):
i appreciate the cleanup! 

i would ask that you review other providers and get an understanding of how they work a bit more - the idea is that anyone can add another implementation at any time, just by extending the enum type of the `backend` option

the `scheduler` provider can serve as a good example here

aanderse (approved):
thank you so much!

PR #127, opened 2026-07-07
feat(console): add environment variables configurations

FixeQD:
Ported standard NixOS options for global session variables to finix. It implements `environment.variables` via `profile.d/session-vars.sh`.

### DIFF
--- a/modules/environment/shells/default.nix
+++ b/modules/environment/shells/default.nix
@@ -4,10 +4,54 @@
   utils,
   ...
 }:
+let
+  atom = lib.types.oneOf [
+    lib.types.int
+    lib.types.str
+    lib.types.path
+  ];
+
+  toStr = v: if lib.isPath v then "${v}" else toString v;
+
+  sessionVarsScript = lib.concatStringsSep "\n" (
+    lib.mapAttrsToList (name: value: "export ${name}=${lib.escapeShellArg value}") (
+      lib.filterAttrs (_: value: value != null) config.environment.variables
+    )
+  );
+in
 {
   options.environment.shells = lib.mkOption {
     type = with lib.types; listOf (either shellPackage path);
     default = [ ];
+  };
+
+  options.environment.variables = lib.mkOption {
+    type = lib.types.attrsOf (
+      lib.types.nullOr (lib.types.coercedTo atom lib.singleton (lib.types.listOf atom))
+    );
+    default = { };
+    apply = lib.mapAttrs (
+      _: value: if value == null then null else lib.concatMapStringsSep ":" toStr value
+    );
+    description = ''
+      Environment variables to set for POSIX-compliant login shells (bash, zsh,
+      dash, ...) via {file}`/etc/profile.d/`. Shells with non-POSIX syntax,
+      such as `fish`, do not source these scripts and will not pick up these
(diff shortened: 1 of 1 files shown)


modules/environment/shells/default.nix
```diff
@@ -6,13 +6,36 @@
...
   utils = import (pkgs.path + "/nixos/lib/utils.nix") { inherit config pkgs lib; };
+
+  sessionVarsScript = lib.concatStringsSep "\n" (
+    lib.mapAttrsToList (
+      name: value:
+      let
+        v = if lib.isList value then lib.concatStringsSep ":" value else value;
+      in
+      "export ${name}=${lib.escapeShellArg v}"
+    ) config.environment.sessionVariables
```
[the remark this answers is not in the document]
FixeQD: This is a static system configuration file compiled from local Nix expressions. If an attacker has arbitrary control over the attribute keys inside my own system configuration, the entire machine is already fully compromised long before this script is even generated. Applying a shell identifier regex here to prevent "shell injection" against myself makes zero sense 😭🙏

modules/environment/shells/default.nix
```diff
@@ -6,13 +6,36 @@
...
+  );
 in
 {
   options.environment.shells = lib.mkOption {
     type = with lib.types; listOf (either shellPackage path);
     default = [ ];
   };
 
+  options.environment.sessionVariables = lib.mkOption {
```
aanderse: in `nixos` this option is called `environment.variables`, not `environment.sessionVariables`.

modules/environment/shells/default.nix
```diff
@@ -6,13 +6,36 @@
...
+    type = with lib.types; attrsOf (either str (listOf str));
+    default = { };
+    description = ''
+      Environment variables to set for all login shells via /etc/profile.d/.
```
aanderse: _all_ login shells? what about `fish` shell? and other shells which aren't `posix`?

modules/environment/shells/default.nix
```diff
@@ -6,13 +6,36 @@
...
+  );
 in
 {
   options.environment.shells = lib.mkOption {
     type = with lib.types; listOf (either shellPackage path);
     default = [ ];
   };
 
+  options.environment.sessionVariables = lib.mkOption {
+    type = with lib.types; attrsOf (either str (listOf str));
```
aanderse: should we allow `null` here to cancel out set values? are we limiting to `str`? it definitely makes sense to include `path` as well... maybe even `int` and convert all to `str`?

you should consider using something like `lib.types.coercedTo atom lib.singleton (lib.types.listOf atom))` where `atom` is your singular data type (`str` currently, though potentially `with lib.types; oneOf [ int str path ];` - see how the `pam` module handles this

modules/environment/shells/default.nix
aanderse: when should we prefer this option over `security.pam.environment`? when should we prefer `security.pam.environment`?

aanderse (commented):
if you don't mind could you please split these into 2 separate PRs? 🙇‍♂️

i have capacity to review the shell variable stuff at the moment, i can return to the console stuff after...

PR #129, opened 2026-07-07
providers/privileges: updates to how providers work for privileges.

xZecora:


aanderse:
what is the specific use case you have for adding these options to the `privileges` provider? i'm trying to get a little bit more context before we extend its abilities

xZecora:
The main option I have a use case for is persistent credentials. It just allows the specific usage of the provider to persist the credentials of the user for some predefined amount of time. `doas` provides this capability per option, `sudo` only allows a global setting for a user/group.

As for the `keepEnv` and `keepEnvVars` options, this just allows you to specify if the escalation retains the entire environment or just a small amount or none at all. The user may want to hide some credentials from specific commands for privacy or security reasons. Both `doas` and `sudo` support this so it seemed simple to add.

There are likely other common options for privilege providers that could be added that people with more complex setups may use, specifically for multi user systems. Those may fall more under the umbrella of a community module because of the sheer amount, however.

As for the options added to specific programs outside of their providers, those should be moved to another PR if they're wanted upstream, they were just part of this branch when I pushed it for you to take a look at my ideas.

PR #131, opened 2026-07-09
modules/services/udev: include eudev's bundled default rules

N1meses:
Issue: if udev is used without elogind it is missing crucial rules. This manifests as: no kernel module autoload for hotplugged devices, no /dev/disk/by-* symlinks after switch-root, missing default device permissions/tags.

Fix: include the default rules in the aggregation.

Build and tested on hardware


aanderse (approved):
thank you! 🙇‍♂️

PR #133, opened 2026-07-09
(fix) modules/services/networkmanager: set rc-manager explicitly

N1meses:
Issue: NetworkManager get build in nixpkgs with a hardcoded resolfconf and rc-manager="auto". If /modules/programs/resolveconf never gets enabled it never gets configured and silently fails.  

Fix:  set rc-manager explicitly based on if  programs.resolvconf.enable then it  contributes its entries via resloveconf  else it symlinks its own resolveconf.

Tested on hardware, in different configurations:
 - standalone NM : /etc/resolv.conf written by  NM, resolution works 
 - NM with programs.resolvconf.enable: NM submits via openresolv and the file is generated by resolvconf

### DIFF
--- a/modules/services/networkmanager/default.nix
+++ b/modules/services/networkmanager/default.nix
@@ -43,6 +43,13 @@ in
     services.dbus.packages = packages;
     services.udev.packages = packages;
 
+    # nixpkgs builds NetworkManager with a hardcoded resolvconf path, so the
+    # default rc-manager=auto always selects resolvconf, even when nothing
+    # has set it up, silently dropping DNS updates.
+    environment.etc."NetworkManager/NetworkManager.conf".text = lib.generators.toINI { } {
+      main.rc-manager = if config.programs.resolvconf.enable then "resolvconf" else "symlink";
+    };
+
     finit.services.network-manager = {
       description = "network manager service";
       conditions = "service/dbus/ready";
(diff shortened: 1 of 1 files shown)


modules/services/networkmanager/default.nix
```diff
@@ -43,6 +43,13 @@ in
     services.dbus.packages = packages;
     services.udev.packages = packages;
 
+    # nixpkgs builds NetworkManager with a hardcoded resolvconf path, so the
+    # default rc-manager=auto always selects resolvconf, even when nothing
+    # has set it up, silently dropping DNS updates.
+    environment.etc."NetworkManager/NetworkManager.conf".text = lib.generators.toINI { } {
```
aanderse:
```suggestion
    environment.etc."NetworkManager/conf.d/00-nixos.conf".text = lib.generators.toINI { } {
```

[upstream documentation](https://man.archlinux.org/man/NetworkManager.conf.5) suggests we add a file to `conf.d/` instead, let's do that

aanderse (commented):
awesome! thank you for contributing. one small nitpick i think is worth adding, if you don't mind 🙇‍♂️

modules/services/networkmanager/default.nix
N1meses: could also name it 00-finix.conf XD
otherwise will I will apply the suggestion and test it

aanderse: until they come at me with their [trademark](https://discourse.nixos.org/t/announcing-nixos-trademark/78585) i will continue to insist `finix` is basically just `nixos`! :laughing:

i'm indifferent, take your pick

N1meses: well lets keep it at that for now

N1meses:
tests checked out
standalone NM: symlinks resolveconf
resolvconf + NM: resolveconf generate it

PR #135, opened 2026-07-10
Fix luks and lvm dm udev sync

Ollie-Gutierrez:
Turned out to be a very simple fix though a little counter intuitive. This also works on the new keventd module aswell with no changes needed, and will hopefully do the same for any other new device managers.

### DIFF
--- a/modules/filesystems/luks.nix
+++ b/modules/filesystems/luks.nix
@@ -68,17 +68,23 @@
         "aes_generic"
       ];
 
-      boot.initrd.fileSystemImportCommands = lib.mkOrder 500 (
-        lib.concatStringsSep "\n" (
-          lib.mapAttrsToList (
+      boot.initrd.finit.tasks.luks =
+        let
+          devices = lib.filterAttrs (_: fs: fs.fsType == "luks") config.fileSystems;
+        in
+        lib.mkIf (devices != { }) {
+          conditions =
+            lib.optionals config.services.mdevd.enable [ "run/coldplug/success" ]
+            ++ lib.optionals config.services.gardendevd.enable [ "run/gardendevctl:2/success" ]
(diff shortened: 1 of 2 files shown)


Ollie-Gutierrez:
This should close https://github.com/finix-community/finix/issues/78 too. Just realised I forgot to run nixfmt on it aswell 💀

aanderse:
thank you *so* much!!

i will rebase, format, and merge after https://github.com/finix-community/finix/pull/137

🚀

aanderse:
i migrated off the new deprecated `fileSystemImportCommands` in favour of `boot.initrd.finit.tasks` for fully parallel booting :sunglasses:

@Ollie-Gutierrez are you able to test and confirm i didn't cause any regressions? :sweat_smile:

Ollie-Gutierrez:
I've just tested it and it doesn't seem to be working, looks like since everything is happening all at once finit is trying to mount before luks is unlocked. I think this should be easy enough to solve by just adding a condition for luks and lvm opening before attempting to mount, not tried it yet though so we'll see!

aanderse:
ah sorry i didn't mean to cause a hassle... if you don't get a working solution soon we can just go back to your original plan, that can work... i don't want to hold people up, so sorry :bowing_man:

Ollie-Gutierrez:
Okay after a bit more looking it's actually not your changes. The way you did is great with the wait tasks it actually solves an issue if been meaning to address with my luks and lvm commits in that they were ordering specific, whereas now they just wait for the parent to finish mounting. Anyway I think my issue is actually with the preservation fork I'm trying out as it's conditional on mount-all which obviously doesn't exist anymore. Gonna see if I can fix it or just test it without 👍

Ollie-Gutierrez:
Okay that's what it was, this should be good to merge!

aanderse:
thanks so much!

PR #136, opened 2026-07-11
fix(sddm): fixup pam/session stuff to match the other DMs

FixeQD:
Cleaned up sddm's PAM stack and session config to match the rest of the login managers, fixed the elogind hardcode causing unnecessary builds and boot hangs, and added proper Wayland session support

Changes:
1. PAM/elogind: Fixed pam_elogind.so in sddm-greeter being hardcoded instead of gated behind services.elogind.enable and added the missing pam_rundir.so + pam_limits.so lines it was also missing
2. Power commands: Replaced the hardcoded loginctl poweroff/reboot with a branch on services.elogind.enable vs providers.privileges
3. Session directories: Fixed X11.SessionDir only pointing at openbox's own xsessions output instead of the merged system profile path, and added a Wayland settings block since sddm had none before

### DIFF
--- a/modules/services/sddm/default.nix
+++ b/modules/services/sddm/default.nix
@@ -43,8 +43,16 @@ in
 
     services.sddm.settings = {
       General = {
-        HaltCommand = "/run/current-system/sw/bin/loginctl poweroff";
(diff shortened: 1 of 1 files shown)


aanderse (approved):
amazing! thank you so much 🤩

PR #141, opened 2026-07-13
feat(console): add custom TTY font/color configurations

FixeQD:
Ported standard NixOS options for virtual console styling to finix. 
It adds support for loading custom virtual console fonts and TTY color palettes and includes an `earlySetup` option to load console fonts straight inside the initrd stage


aanderse:
future work: replace `boot.initrd.fileSystemImportCommands` which is now deprecated

modules/hardware/console.nix
```diff
@@ -19,6 +19,36 @@ let
...
+        mapArg = lib.optionalString (
+          cfg.keyMap != null
+        ) " -m ${fontEnv}/share/consolefonts/${lib.escapeShellArg cfg.keyMap}.acm 2>/dev/null || true";
+      in
+      "${pkgs.kbd}/bin/setfont ${fontArg} -C /dev/console || ${pkgs.kbd}/bin/setfont ${fontEnv}/share/consolefonts/${fontArg} -C /dev/console${mapArg}";
+
+  colorsScript = lib.optionalString (cfg.colors != [ ]) (
+    let
+      inherit (lib) imap0 concatStringsSep;
+      esc = "\033]P";
```
eclairevoyant, line 46: The `0` here is what gets escaped because of how nix works, so the string will literally contain `033]P`. Was this code actually tested?

aanderse: @FixeQD are you able to comment?

@eclairevoyant is this causing upstream `hjem` issues? i suppose `"\\033]P"` is the proper value here?

eclairevoyant: No issues on hjem's end, just noticed a lot of warnings in the logs and wanted to check. Yeah the `\\` is probably the correct version.

PR #146, opened 2026-07-15
dinit initial work

willowispll:


modules/dinit/default.nix
```diff
@@ -0,0 +1,147 @@
...
+
+        # Stop services that were enabled in the previous generation but are no longer enabled
+        oldServiceDir="/run/current-system/etc/dinit.d"
+        if [ -d "$oldServiceDir" ]; then
+          declare -A enabled_services=( ${enabledAssoc} )
+          for f in "$oldServiceDir"/*; do
+            [ -e "$f" ] || continue
+            name="$(basename "$f")"
+            if [ -z "''${enabled_services[$name]-}" ]; then
+              ${dinitctl} stop --no-wait "$name" 2>&1 | logger -t finix-dinit || true
```
y0usaf: This stops `boot` on every switch. The old dinit.d dir always contains `boot` and `boot.d`, and neither is ever in `enabled_services`, so they both hit this branch.

I went through the dinit source to check what that actually does: `boot` is started with explicit activation, so `dinitctl stop boot` drops its required_by to 0, which releases all of its dependencies — i.e. everything symlinked into boot.d gets stopped too. On a dinit system this takes down the whole service tree on every switch.

Need to skip `boot`/`boot.d` here.

modules/dinit/default.nix
```diff
@@ -0,0 +1,147 @@
...
+        An attribute set of `dinit` system level services.
+
+        See [upstream documentation](https://davmac.org/projects/dinit/man-pages-html/dinit-service.5.html) for additional details.
+      '';
+    };
+  };
+
+  config = {
+    environment.systemPackages = lib.mkIf (cfg.services != { } || cfg.user.services != { }) [
+      pkgs.dinit
```
y0usaf: Should be `cfg.package`, not `pkgs.dinit` — same for the dinitctl path in dinit-reload (line 125) and the `boot.init` fallback in boot/default.nix. As written the `dinit.package` option doesn't actually do anything.

modules/dinit/default.nix
```diff
@@ -0,0 +1,147 @@
...
+    system.activation.scripts.dinit-reload = {
+      deps = [ "etc" "dinitBootD" ];
+      text = let
+        enabledNames = lib.attrNames (lib.filterAttrs (_: s: s.enable) cfg.services);
+        enabledList = lib.concatStringsSep " " (map (n: "\"${n}\"") enabledNames);
+        enabledAssoc = lib.concatMapStringsSep " " (n: "[\"${n}\"]=1") enabledNames;
+        dinitctl = "${pkgs.dinit}/bin/dinitctl";
+      in ''
+        # Reload definitions for services in the new config
+        for svc in ${enabledList}; do
```
y0usaf: Newly-enabled services never get started here (reload only works on already-loaded services), and removed ones get stopped but not unloaded, so stale definitions stick around until reboot. Fine for initial work, but worth a TODO so nobody assumes switch is fully wired up.

modules/dinit/switch-to-configuration.sh
```diff
@@ -0,0 +1,101 @@
...
+"$logger/bin/logger" -t finix "starting switch-to-configuration ($action)"
+
+if [[ "$action" != boot && "${NIXOS_NO_CHECK-}" != 1 ]]; then
+  if ! "$inhibitCheck" "$out"; then
+    exit 1
+  fi
+fi
+
+# Mount filesystems from fstab (needed for efivars during limine install)
+"$coreutils/bin/mount" -a 2>/dev/null || true
```
y0usaf: coreutils doesn't ship a `mount` binary — this fails every time and the `|| true` hides it. The comment says this is needed for efivars during limine install, so it's silently broken rather than optional. Use util-linux.

modules/dinit/switch-to-configuration.sh
```diff
@@ -0,0 +1,101 @@
+#!@bash@/bin/bash
+set -euo pipefail
+
+out="@out@"
+localeArchive="@localeArchive@"
```
y0usaf: `@localeArchive@` never gets substituted — the dinit branch in activation/default.nix doesn't pass it to substituteInPlace, so the built script ships the literal placeholder. Harmless while the usage below is commented out, but either wire it up or drop the var.

modules/system/activation/default.nix
```diff
@@ -192,16 +192,22 @@ in
             ${coreutils}/bin/ln -s ${config.boot.initrd.package}/initrd $out/initrd
           ''
           + ''
-            cp ${../../finit/switch-to-configuration.sh} $out/bin/switch-to-configuration
+            cp ${
+              if config ? dinit && config.dinit.services != { }
```
y0usaf: Since mount-fstab is defined unconditionally in the dinit module, `dinit.services != { }` is true the moment the module is imported at all. End result on a default system: finit is still PID1 (the init symlink), but the dinit stc script gets installed — and the dinit script doesn't do finit's `initctl reload`, so finit services stop reloading on switch. Verified by building a toplevel with just the module imported.

Can this key off `config.boot.init` instead?

modules/dinit/common-options.nix
```diff
@@ -0,0 +1,447 @@
...
+        ]
+      '';
+      description = ''
+        Packages added to the `PATH` environment variable of this service.
+      '';
+    };
+
+    # standard dinit options, applicable to both user and system level services
+
+    type = lib.mkOption {
```
y0usaf, line 42: No default here means omitting `type` gives a confusing 'option accessed but has no value defined' eval error, since the generator forces every attribute. dinit itself defaults to `process` — worth mirroring that with `default = "process"`.

y0usaf (commented):
Evaluated and built this locally — the format layer is solid (yes/no bools, duplicate keys for deps, space-joined options, env-file generation all match dinit-service(5)), and the .keep placeholder in boot.d is safe since dinit ignores dot-entries when scanning dependency dirs.

A few blockers inline, the stop-loop one is the serious one.

Bigger picture, for later: a dinit-booted system currently loses the sysctl/modprobe/remount-nix-store tasks since those are all gated on finit.enable, so having boot.init silently default to dinit when finit is off feels premature. I'd make dinit-as-init an explicit opt-in until equivalents exist. Same story with the nixos-compat tmpfiles shim now being tied to finit.enable.

willowispll:
It seems to work and it also fixes all issues (for now and at least all issues i could think of). Needs better testing tho and a code review.

PR #147, opened 2026-07-15
modules/fish: Added customization options.

Sqydev:
Added options: shellAbbrs, shellAliases, shellInit to fish module so you can have your fish config in your finix config like in nixos.
It works.


aanderse:
thanks for contributing this @Sqydev! i'm trying to figure out how to best abstract (or not!) this functionality behind different shells... i hope to return to this soon...

Issue #161, opened 2026-07-18
retracted — opened in error


y0usaf:
This issue was opened in error. Please disregard; a maintainer may delete it. Apologies for the noise.

aanderse:
i accepted this as a compromise for first revision of the initramfs rewrite, but i would be happy to accept any pull requests correcting the situation here!

y0usaf:
Opened in error — please disregard and delete if possible. Apologies for the noise.

PR #164, opened 2026-07-18
fix(lib): make escapePath injective (systemd-escape convention)

y0usaf:
Follow-up to #161 — you mentioned you'd accept a PR correcting this.

**Problem**: `escapePath "/"` and `escapePath "/root"` both produce `"root"`. With a `neededForBoot` bind mount of `/root`, the two generated mount stanzas collide, and `genAttrs'` silently keeps one — the root filesystem's mount task is overwritten by the `/root` bind and never runs.

Reproducer (any system with `/` + a `/root` bind both `neededForBoot`):

```nix
fileSystems."/" = { device = "none"; fsType = "tmpfs"; neededForBoot = true; };
fileSystems."/persist" = { device = "/dev/sda2"; fsType = "btrfs"; neededForBoot = true; };
fileSystems."/root" = { device = "/persist/root"; fsType = "btrfs"; options = [ "bind" ]; neededForBoot = true; };
```

Generated `boot.initrd.finit.tasks` on main:
`[ "mount-persist" "mount-root" "wait-dev-persist" ]` — the `/` mount task is silently lost.

**Fix**: adopt the `systemd-escape --path` convention — `/` escapes to `-`, everything else keeps the current scheme (leading slash stripped, `/` → `-`). Injective, and every non-root path keeps its existing stanza name; only the root fs stanza is renamed (`mount-root` → `mount--`). All conditions are generated through the same function, so they stay consistent.

Same config on this branch:
`[ "mount--" "mount-persist" "mount-root" "mount-var-lib-foo" "wait-dev-persist" ]`, with parent conditions correctly referencing `task/mount--/success`.

### DIFF
--- a/modules/lib/utils.nix
+++ b/modules/lib/utils.nix
@@ -2,8 +2,15 @@
 { lib }:
 {
   # simple path escape into a name safe for use as a finit stanza name and conditions
+  #
+  # Mapping (systemd-escape --path convention): "/" becomes "-", and any
+  # other path has its leading slash stripped and remaining slashes
+  # replaced with "-". This is injective: "/" used to escape to "root",
+  # which collided with "/root" (e.g. a neededForBoot bind mount of /root
+  # tripped the mount.nix uniqueness assert); every non-root path keeps
+  # the name it had before.
   escapePath =
-    s: if s == "/" then "root" else lib.replaceStrings [ "/" ] [ "-" ] (lib.removePrefix "/" s);
+    s: if s == "/" then "-" else lib.replaceStrings [ "/" ] [ "-" ] (lib.removePrefix "/" s);
 
   # Convert a shell package or path into an absolute shell path.
   toShellPath =
--- a/modules/lib/utils.nix
+++ b/modules/lib/utils.nix
@@ -3,14 +3,16 @@
 {
   # simple path escape into a name safe for use as a finit stanza name and conditions
   #
-  # Mapping (systemd-escape --path convention): "/" becomes "-", and any
-  # other path has its leading slash stripped and remaining slashes
-  # replaced with "-". This is injective: "/" used to escape to "root",
-  # which collided with "/root" (e.g. a neededForBoot bind mount of /root
-  # tripped the mount.nix uniqueness assert); every non-root path keeps
-  # the name it had before.
+  # "/" escapes to "" — call sites compose "<prefix>-${escapePath p}", so the
+  # root filesystem gets the bare prefix (e.g. "mount-") and every other path
+  # keeps its previous name (leading slash dropped, "/" -> "-"). "/" used to
+  # escape to "root", which collided with "/root" and tripped the mount.nix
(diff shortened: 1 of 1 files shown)


N1meses:
Now this send me down a rabbit hole.
This doesn't work on finit beacuse "-- " is treated as a description delimiter which renders everything after it as a descrition. https://github.com/finit-project/finit/blob/master/src/service.c line:1897.

In of itself this isn't an issue because the name is then "mount--" which is missing the space.
But then in https://github.com/finix-community/finix/blob/main/modules/finit/stage1.nix
specifically: 
```nix
  serviceStr =
    svcType: svc:
    lib.concatStringsSep " " (
      [
        svcType
        "[${svc.runlevels}]"
      ]
      ++ lib.optional (svc.name or null != null) "name:${svc.name}"
      ++ lib.optional (svc.id or null != null) ":${svc.id}"
      ++ lib.optional (svc.respawn or false) "respawn"
      ++ lib.optional (svc.restart or null != null) "restart:${toString svc.restart}"
      ++ lib.optional (svc.notify or null != null) "notify:${svc.notify}"
      ++ lib.optional (svc.conditions or [ ] != [ ]) "<${lib.concatStringsSep "," svc.conditions}>"
      ++ lib.optional (svc.tty or null != null) "tty:${svc.tty}"
      ++ lib.optional (svc.extraConfig or "" != "") svc.extraConfig
      ++ lib.optional (svc.command or null != null) svc.command
      ++

        # tty specific options
        (lib.optional (svc.device or null != null) svc.device)
      ++ lib.optional (svc.baud or null != null) svc.baud
      ++ lib.optional (svc.noclear or false) "noclear"
      ++ lib.optional (svc.nowait or false) "nowait"
      ++ lib.optional (svc.nologin or false) "nologin"
      ++ lib.optional (svc.rescue or false) "rescue"
      ++ lib.optional (svc.notty or false) "notty"
      ++

        (lib.optional (svc.description != null) "-- ${svc.description}")
    );
```
concatenates everything to a single line string, separated by a " " which makes everything in the mount root task after "name:mount-- " a description and cuts it when parsed at boot. So task/mount--/success is never emitted and hangs

TL;DR: "--" is unsafe on finit

y0usaf:
Good catch, thanks for chasing it into service.c. Switched the escape to `""` instead — call sites already end in `-`, so the root task is just `mount-` and `--` can't occur. A path that itself contains `--` would still trip this, but that's pre-existing; probably worth a separate issue.

N1meses:
Thank you for making changes and i tested it one 2 devices with no issues there, everything booted fine.

Tho now things changed, with #181 and the new syntax for finit, "--" could work as it is no longer used for the description delimiting. Now it follows a key-value format tho i have to do more reading to actually be sure. 

I think @aanderse should weigh in on this.

aanderse:
good point about `finit v5`! if we don't have any more restrictions with `--` what is the best escape algorithm we can follow? i guess we should take a look at what is all possible? do we want to just follow what `systemd` does here, if legal, or someone more straightforward?

PR #166, opened 2026-07-21
modules/services/mdevd: prepend kmod so its modprobe isn't shadowed by busybox

N1meses:
Issue: With programs.coreutils.package = pkgs.busybox, mdevd's modalias-based autoload silently fails: no input (evdev), wifi, or touchpad, while by-name module loading and the console keyboard still work.

Fix: put pkgs.kmod ahead of coreutils in the path.

tested on: hardware with mdvd, seatd, busybox

### DIFF
--- a/modules/services/mdevd/default.nix
+++ b/modules/services/mdevd/default.nix
@@ -45,7 +45,7 @@ let
 
   # Insert modules for devices with a modalias.
   # Use @ prefix to run via /bin/sh on add events.
-  modaliasRule = ''-$MODALIAS=.* 0:0 660 @modprobe --quiet "$MODALIAS"'';
+  modaliasRule = ''-$MODALIAS=.* 0:0 660 @${pkgs.kmod}/bin/modprobe --quiet "$MODALIAS"'';
 
   # We need symlinks in /dev/disk/{by-id,by-label,by-uuid,by-partlabel,by-partuuid}
   # so we run this script for block device events.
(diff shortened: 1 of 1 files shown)


modules/services/mdevd/default.nix
```diff
@@ -194,9 +194,9 @@ in
 
       # TODO: now we're hijacking `env` and no one else can use it...
       path = [
+        pkgs.kmod
```
aanderse: `(lib.hiPrio pkgs.kmod)` will be more robust over time... though in this case what you have is also absolutely going to work

worth adding a comment like `# ensure kmod wins over any other packages added to the $PATH of this service`

aanderse (changes requested):
my biggest question - why does `modprobe` from `busybox` not work here 🤔 i'm guessing `pkgs.busybox` isn't [patched to handle the correct paths](https://github.com/NixOS/nixpkgs/blob/fd1462031fdee08f65fd0b4c6b64e22239a77870/pkgs/os-specific/linux/kmod/default.nix#L84)... do we have any desire to "fix" this upstream in `nixpkgs` as an added bonus? maybe not... 🤷‍♂️

modules/services/mdevd/default.nix
N1meses: Or should i rather hardcode the path in modaliasRule?
````
modaliasRule = ''-$MODALIAS=.* 0:0 660 @${pkgs.kmod}/bin/modprobe --quiet "$MODALIAS"'';

aanderse (approved):
wonderful! thank you very much 🚀

PR #167, opened 2026-07-23
feat(i18n): support non-glibc pkgs

adamcstephens:
allow building a musl system.


aanderse (approved):
really happy to see you here @adamcstephens! please don't hesitate to ping me if you need a hand with anything, have any questions, etc...

btw @chocoblocko9 has been running `finix` on his desktop with `wayland` and `musl` for a while now, so i'm sure he appreciates this _proper_ fix 🙂

PR #171, opened 2026-07-25
modules/filesystems/zfs: fix boot hang on native datasets

N1meses:
Issue: zfs native datasets get imported and automatically mounted at their native mount points in stage 1. this conflicts with stage 1's /nix if one has tank/nix and causes hang on boot.

fix: use the -N flag which prevents the datasets from being auto mounted in stage 1.  the dataset tank/nix would then get mounted at /sysroot/nix which becomes /nix after switch-root and with that the correct mount point for the dataset.

tested on zfs device
in theory legacy-datasets are unaffected since they never auto mount to begin with and -N would no-op for them. but i would rather have it tested then guess

### DIFF
--- a/modules/filesystems/zfs.nix
+++ b/modules/filesystems/zfs.nix
@@ -113,7 +113,7 @@
                 tries=900
                 i=0
                 while [ "$i" -lt "$tries" ]; do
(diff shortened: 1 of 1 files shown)


aanderse:
good catch! i only use legacy datasets! 

i will test with legacy datasets this weekend. thanks so much!

PR #172, opened 2026-07-25
modules/services/docker: fix reference to non-existent option config.boot.zfs.package

N1meses:
issue: if one runs docker and zfs, it references a non existent i option.boot.zfs.packages. this fails at eval
(pobably from nixos module copied over)

fix: reference config.boot.supportedFilesystems.zfs.packages instead

tested; will eval now and still show Storage Driver: zfs


aanderse:
cc @deathbymanatee

PR #173, opened 2026-07-26
ifupdown-ng and dhcpcd: Fix /etc/resolv.conf generation

deinferno:
In imperative distros installing dhcpcd creates /etc/dhcpcd.conf file with useful defaults like querying nameservers. This change makes dhcpcd module render that file for cases where dhcpcd used as dependency.

Second commit makes ifupdown-ng dependent on dhcpcd for rendering this file and by default switches off finit's dhcpcd service to not be in conflict with each other.

Also checkout https://github.com/NixOS/nixpkgs/pull/536753 for nixpkgs ifupdown-ng update.

### DIFF
--- a/modules/programs/ifupdown-ng/default.nix
+++ b/modules/programs/ifupdown-ng/default.nix
@@ -1,4 +1,5 @@
 {
+  modules,
   config,
   pkgs,
   lib,
@@ -14,6 +15,9 @@ let
   multiValueType = with lib.types; nullOr (coercedTo str lib.singleton (listOf str));
 in
 {
+
+  imports = [ modules.dhcpcd ];
+
   options.programs.ifupdown-ng = {
     enable = lib.mkOption {
       type = lib.types.bool;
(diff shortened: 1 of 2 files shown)


aanderse:
i'm using `ifupdown-ng` for static ip assignment only at the moment... seems like my usecase is extremely minimal and doesn't cover much of the functionality... :sweat:

so in most usecases other than mine `ifupdown-ng` has a _hard_ requirement on the `dhcpcd` package and its configuration? but at the same time doesn't want you to run the software... just configure it and have the package available?

modules/programs/ifupdown-ng/default.nix
```diff
@@ -158,6 +162,11 @@ in
   };
 
   config = lib.mkIf cfg.enable {
+
+    # Needed for /etc/dhcpcd.conf
+    services.dhcpcd.enable = true;
+    finit.services.dhcpcd.enable = lib.mkDefault false;
```
aanderse, line 168: i'm not specifically a fan of this... but i'll think it through to see if there is a better alternative...

aanderse:
@deinferno are we waiting on the upstream `nixpkgs` PR?

Issue #105, opened 2026-06-17
More Services


indium114:
Following up on the [Discourse topic](https://discourse.nixos.org/t/init-system-freedom-on-nixos/78333/11)

I'd love to switch to Finix, but there are some `services.*` options that I require, which are missing on Finix.

Namely:
- `services.printing`
- `services.avahi`
- `services.locate`
- `services.displayManager`
- `services.yarr`
- `services.tailscale`
- `services.yggdrasil`
- `services.i2pd`
- `services.kanata`

If any of these already exist, lmk.

aanderse:
users have already contributed [printing](https://github.com/finix-community/community-modules/blob/main/modules/services/cups/default.nix) and [tailscale](https://github.com/finix-community/community-modules/blob/main/modules/services/tailscale/default.nix) modules to the [community-modules](https://github.com/finix-community/community-modules) repo, and we have a `greetd` module for `services.displayManager`... i think some of these other modules are very possible to be added soon

i think we can add a `WISHLIST.md` to the `finix/community-modules` and keep track of user requests

indium114:
> users have already contributed [printing](https://github.com/finix-community/community-modules/blob/main/modules/services/cups/default.nix) and [tailscale](https://github.com/finix-community/community-modules/blob/main/modules/services/tailscale/default.nix) modules to the [community-modules](https://github.com/finix-community/community-modules) repo

I hadn't checked the community repo, thanks for the heads up!

I agree that a `WISHLIST.md` would be a good idea

PR #175, opened 2026-07-28
feat(services): add avahi module

ar175-lol:
Adds `services.avahi` module. 

Tested, with this configuration:
```nix
services.avahi = {
  enable = true;
  hostName = "finixos";
  domainName = "local";
  browseDomains = [ "0pointer.de" "zeroconf.org" ];
  ipv4 = true;
  ipv6 = false;
  wideArea = false;
};
```
Nix generates this config:
```ini
[server]
host-name=finixos
browse-domains=0pointer.de, zeroconf.org
use-ipv4=yes
use-ipv6=no

domain-name=local
allow-point-to-point=no

[wide-area]
enable-wide-area=no

[publish]
disable-publishing=yes
disable-user-service-publishing=yes
publish-addresses=no
publish-hinfo=no
publish-workstation=no
publish-domain=no

[reflector]
enable-reflector=no
```

Relates to #105.

### DIFF
--- a/modules/services/avahi/default.nix
+++ b/modules/services/avahi/default.nix
@@ -1,0 +1,112 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.avahi;
+
+  mkValueString =
+    v:
+    if v == true then
+      "yes"
+    else if v == false then
+      "no"
+    else
+      lib.generators.mkValueStringDefault { } v;
+
+  format = pkgs.formats.ini {
+    mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
+    listToValue = lib.concatMapStringsSep ", " mkValueString;
+  };
+
+  enableDbus = cfg.settings.server.enable-dbus or true;
+in
+{
+  options.services.avahi = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable [avahi](${pkgs.avahi.meta.homepage}) as a system service.
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.avahi;
+      defaultText = lib.literalExpression "pkgs.avahi";
+      description = ''
+        The package to use for `avahi`.
+      '';
+    };
+
+    debug = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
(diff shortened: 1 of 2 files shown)


modules/services/avahi/default.nix
```diff
@@ -0,0 +1,112 @@
...
+    environment.etc."avahi/avahi-daemon.conf".source = format.generate "avahi-daemon.conf" cfg.settings;
+
+    # TODO: add finit.services.reloadTriggers option
+    environment.etc."finit.d/avahi-daemon.conf".text = lib.mkAfter ''
+
+      # reload trigger
+      # ${config.environment.etc."avahi/avahi-daemon.conf".source}
+    '';
+
+    finit.services.avahi-daemon = {
```
aanderse, line 89: do we need to set `path`?

```
path = [
  config.programs.coreutils.package
  cfg.package
];
```

and if we do need it... what do we need it for?

aanderse (commented):
i made a few changes... 😅 do you mind letting me know what you think? if you like, could you test to ensure no regressions?

thank you! 🙇‍♂️

modules/services/avahi/default.nix
ar175-lol: It looks we don't need it. When testing, I didn't hit any issues about `path`. I can test your changes tomorrow 

> do you mind letting me know what you think? 

Looks good! That makes code much cleaner and future-proof, thank you!

ar175-lol:
I tested, with this config:

```nix
services = {
  avahi = {
    enable = true;
    debug = true;
    settings = {
      server = {
        host-name = "finixos";
        domain-name = "local";
        browse-domains = [ "0pointer.de" "zeroconf.org" ];
      };
      publish = {
        disable-publishing = true;
        disable-user-service-publishing = true;
        publish-address = false;
      };
    };
  };
};
```
It is generating this:
image


~I can't check if `avahi`even works, because I have some issues with user and root passwords on my finix VM :P~

The only issue was that it wasn't clear whether avahi was running after the rebuild (avahi is exiting 1 when config is incorrect). Maybe we could add a warning like in NixOS (warning: the following service failed: service-name)? Idea for another PR :)

aanderse:
so everything is working? good to merge?

ar175-lol:
> so everything is working? good to merge?

I think yes

aanderse (approved):
thanks so much for the contribution! i hope you are enjoying finix so far

PR #179, opened 2026-07-29
modules/services/zram-swap: init

y0usaf:
Adds zram swap as an idempotent finit task (same approach as loadkmap, #170): modprobe, skip if a zram device already swaps, then zramctl/mkswap/swapon. Options: enable, memoryPercent, algorithm, priority — defaults match NixOS's zramSwap. Been running this on two machines since mid-July; kept to the common case, no writeback or multi-device.

### DIFF
--- a/modules/services/zram-swap/default.nix
+++ b/modules/services/zram-swap/default.nix
@@ -1,0 +1,73 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.zram-swap;
+in
+{
+  options.services.zram-swap = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable swap on a compressed zram block device.
+      '';
+    };
+
+    memoryPercent = lib.mkOption {
+      type = lib.types.int;
+      default = 50;
+      description = ''
+        Size of the zram device as a percentage of total memory.
+      '';
+    };
+
+    algorithm = lib.mkOption {
+      type = lib.types.str;
+      default = "zstd";
+      example = "lz4";
+      description = ''
+        Compression algorithm passed to {command}`zramctl`.
+      '';
+    };
+
+    priority = lib.mkOption {
+      type = lib.types.int;
+      default = 5;
+      description = ''
+        Swap priority of the zram device. Higher values are preferred by
+        the kernel over lower-priority swap devices.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    boot.kernelModules = [ "zram" ];
+
+    finit.tasks.zram-swap = {
+      description = "zram swap (${toString cfg.memoryPercent}% RAM, ${cfg.algorithm})";
+      log = true;
+      command = pkgs.writeShellScript "zram-swap" ''
+        set -eu
+        export PATH=${
+          lib.makeBinPath [
+            pkgs.coreutils
+            pkgs.util-linux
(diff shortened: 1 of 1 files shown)


modules/services/zram-swap/default.nix
```diff
@@ -0,0 +1,73 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    boot.kernelModules = [ "zram" ];
+
+    finit.tasks.zram-swap = {
+      description = "zram swap (${toString cfg.memoryPercent}% RAM, ${cfg.algorithm})";
+      log = true;
+      command = pkgs.writeShellScript "zram-swap" ''
```
FixeQD, line 60: nit but this task has no `conditions` at all, every other finit task in the repo sets at least `conditions = "service/syslogd/ready"`. without it this can fire before `syslogd` is even up, so `log = true` has nowhere to write to. 

U probably want:
```nix
conditions = [ "service/syslogd/ready" "task/modprobe/success" ];
```

modules/services/zram-swap/default.nix
```diff
@@ -0,0 +1,73 @@
...
+        export PATH=${
+          lib.makeBinPath [
+            pkgs.coreutils
+            pkgs.util-linux
+            pkgs.gnugrep
+            pkgs.gawk
+            pkgs.kmod
+          ]
+        }
+        modprobe zram || true
```
FixeQD: this feels like a workaround for the missing conditions above tbh
Finix already has `boot.kernelModules` -> the central `finit.tasks.modprobe` handles loading everything, same pattern zerotierone/docker use

if you add `task/modprobe/success` to conditions this line can just go away

modules/services/zram-swap/default.nix
```diff
@@ -0,0 +1,73 @@
...
+      '';
+    };
+
+    memoryPercent = lib.mkOption {
+      type = lib.types.int;
+      default = 50;
+      description = ''
+        Size of the zram device as a percentage of total memory.
+      '';
+    };
```
FixeQD, line 29: minor: `memoryPercent` as a plain int accepts 0 or >100, which only fails later at `zramctl` runtime

It's worth noting upstream nixpkgs has the exact same gap and has open issues about weird behavior at the edges (NixOS/nixpkgs#103106), so this isn't a blocker, just something to maybe do better than they did

FixeQD (changes requested):
looks solid overall, left a couple comments, the conditions one is the only thing I'd actually block on, rest is take-it-or-leave-it :D

modules/services/zram-swap/default.nix
y0usaf: Addressed in 0367f49: the task now waits for service/syslogd/ready and task/modprobe/success. Verified the evaluated finit task conditions.

modules/services/zram-swap/default.nix
y0usaf: Addressed in 0367f49: removed the redundant modprobe invocation and kmod PATH dependency; boot.kernelModules and the central modprobe task handle loading.

modules/services/zram-swap/default.nix
y0usaf: Addressed in 0367f49 with lib.types.ints.positive: zero and negative values now fail evaluation. Values above 100 remain supported because this is the logical size of a compressed device, not a physical RAM allocation; documented the compressibility caveat. Verified -1/0 rejection and 50/150 acceptance.

FixeQD (approved):
LGTM

PR #180, opened 2026-07-29
modules/services/tailscale: init

y0usaf:
One of the modules requested in #105. Shape follows zerotierone (the closest existing module, also a mesh VPN): enable/package/port/stateDir/extraFlags, tun kernel module, finit-supervised daemon. iproute2 and iptables sit on the service path because tailscaled shells out to both. Running on my server and desktop since mid-July.

### DIFF
--- a/modules/services/tailscale/default.nix
+++ b/modules/services/tailscale/default.nix
@@ -1,0 +1,233 @@
+{
+  config,
+  pkgs,
+  lib,
+  options,
+  ...
+}:
+let
+  cfg = config.services.tailscale;
+
+  tun = cfg.interfaceName != "userspace-networking";
+
+  routingSysctls =
+    lib.optionalAttrs
+      (builtins.elem cfg.routingSysctls [
+        "server"
+        "both"
+      ])
+      {
+        "net.ipv4.conf.all.forwarding" = lib.mkDefault true;
+        "net.ipv6.conf.all.forwarding" = lib.mkDefault true;
+      }
+    //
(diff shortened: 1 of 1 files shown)


ar175-lol:
I noticed you set `log = true;` in the service configuration, can you please tell me what does this do? Thanks! (I am new to Finix, just started exploring it a couple days ago)

aanderse:
> allow the keyword `log` to redirect `stderr` and `stdout` of the application to a file or syslog

https://finit-project.github.io/config/logging/?h=log#redirecting-output

willowispll:
@y0usaf have you seen https://github.com/finix-community/community-modules/blob/main/modules/services/tailscale/default.nix?

it has more features and was already tested, consider adapting them 

cc @parzivale

y0usaf:
Adapted the feature set from the community module into this PR — thanks @willowispll. Kept core option style (long-form mkOption, zerotierone shape), iproute2/iptables on the service path, and the stateDir tmpfiles rule guarded on the default per its docstring.

willowispll:
@aanderse why it is not merged yet? 
looks good to me tho i have not tested it

y0usaf:
I can clank away and shape it in any shape you want :') just let me know

willowispll:
it looks good enough for me? 
i dont think we need to change something

aanderse:
i did a bit of research or `tailscale` a while ago and found some room for improvement/standardization that i was hoping to get to... let me try to find my notes and get back to you on this

PR #181, opened 2026-08-01
migrate to finit v5

aanderse:


modules/finit/options.nix
```diff
@@ -0,0 +1,799 @@
...
+    extraPackages = lib.unique (
+      lib.flatten (
+        lib.concatMap (v: lib.optional v.enable (v.packages or [ ])) (
+          lib.attrValues config.boot.supportedFilesystems
+        )
+      )
+    );
+  };
+
+  execType = with lib.types; nullOr (coercedTo path toString (coercedTo program lib.getExe str));
```
N1meses: odrdering should be:

```suggestion
  execType = with lib.types; nullOr (coercedTo program lib.getExe (coercedTo path toString str));
```

N1meses (changes requested):
i think this is the issue

modules/finit/options.nix
aanderse: just pushed a fix, thanks!

PR #184, opened 2026-08-01
boot/initrd: split boot.initrd.path across owning modules

FixeQD:
Resolves the TODO in modules/boot/initrd.nix ("split this up, allow modules to contribute what they need"). So `boot.initrd.path` no longer hardcodes a fixed package list

- modules/boot/initrd.nix: keeps only `pkgs.busybox` + the already-modular `fsPackages`
- modules/finit/mount.nix: contributes `hiPrio pkgs.util-linux.mount`, since it's the module that calls bare `mount`
- modules/boot/modprobe.nix: contributes `pkgs.kmod`, gated by the same cfg.enable as the rest of that module
- pkgs.bash dropped entirely - nothing in the initrd invokes it bare, all scripts run via `#!/bin/sh` (`busybox`) or absolute store paths


modules/boot/modprobe.nix
```diff
@@ -19,6 +19,8 @@ in
   };
 
   config = lib.mkIf cfg.enable {
+    boot.initrd.path = [ pkgs.kmod ];
+
```
aanderse: `kmod` is unconditionally required by the initramfs

aanderse (changes requested):
*very* nice! thank you! just one minor change please 🙇‍♂️

modules/boot/modprobe.nix
```diff
@@ -18,34 +18,42 @@ in
...
-      '';
-    };
+    (lib.mkIf cfg.enable {
+      environment.etc."modules-load.d/finix.conf" = {
+        text = ''
+          set modprobe = ${pkgs.kmod}/bin/modprobe
 
-    environment.etc."modprobe.d/ubuntu.conf".source = "${pkgs.kmod-blacklist-ubuntu}/modprobe.conf";
-    environment.etc."modprobe.d/debian.conf".source = pkgs.kmod-debian-aliases;
+          ${lib.concatStringsSep "\n" config.boot.kernelModule3s}
```
N1meses: typo in kernelModules

aanderse (approved):
did anyone test this yet? looks great!

N1meses (changes requested):
sorry i did a review but forgot to comment 
still kind of new to working on a project like this
just this one which is a blocker

modules/boot/modprobe.nix
FixeQD: Oops

N1meses (approved):
Works fine for me and boots on all 4 finix hosts i have

PR #185, opened 2026-08-01
dinit pr

willowispll:


willowispll:
tested it on my machine, boots correctly, services are also correctly symlinked to targets

PR #186, opened 2026-08-01
modules/mango: replace hardcoded wlroots version with reflection

Carrot-shreds:
Nixpkgs-unstable bumped `mango` version to `0.15.5` last week https://github.com/NixOS/nixpkgs/pull/539969
, and now it use `wlroots_0_20`.
Previously hardcoded wlroots version will cause build failed and need more maintenance. Now it doesn't relay upstream version any more.
I only tested it on my laptop using latest nixpkgs-unstable. But it should also keep backward compatibility when using nixpkgs 26.05 or 25.11 with `mangowc` and `wlrooots_0_19`.

### DIFF
--- a/modules/programs/labwc/default.nix
+++ b/modules/programs/labwc/default.nix
@@ -45,12 +45,17 @@ in
 
     package = lib.mkOption {
       type = lib.types.package;
-      default = pkgs.labwc.override {
-        inherit libinput;
-
-        wlroots_0_20 = pkgs.wlroots_0_20.override { inherit libinput; };
-        enableSystemd = false;
-      };
+      default = pkgs.labwc.override (
+        o:
+        let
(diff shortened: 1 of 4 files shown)


aanderse:
thanks for contributing! we have a couple programs that depend on `wlroots` so i took the liberty of adding to your PR... :sweat_smile:

could you please let me know if you like this and maybe give it a quick test? :bowing_man:

Carrot-shreds:
I don't have much experience about nix language, thank you for making the code more simple and readable!
I just tested all these four modules, and also fixed two small problems.
- labwc: correct override attr `withSystemd`to `enableSystemd`.
- regreet: mv `enableXWayland` override attr from `cage` to `wlroots`.

aanderse:
cc @deathbymanatee updates to `wlroots` shouldn't be a problem anymore :rocket:

thanks again @Carrot-shreds! i hope you are enjoying `finix` :smile:

PR #187, opened 2026-08-01
modules/services/yarr: init

ar175-lol:
This PR adds `services.yarr` option. Tested on finix VM, service is running properly.

Related to #105

### DIFF
--- a/modules/services/yarr/default.nix
+++ b/modules/services/yarr/default.nix
@@ -1,0 +1,95 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
+  cfg = config.services.yarr;
+in {
+  options.services.yarr = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable [yarr](${pkgs.yarr.meta.homepage}) as a system service.
+      '';
+    };
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.yarr;
+      description = ''
+        The package to use for `yarr`.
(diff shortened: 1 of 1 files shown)


modules/services/yarr/default.nix
```diff
@@ -0,0 +1,95 @@
...
+      description = "Base path of the service url.";
+    };
+
+    authFilePath = lib.mkOption {
+      type = lib.types.nullOr lib.types.path;
+      default = null;
+      description = "Path to a file containing username:password. `null` means no authentication required to use the service.";
+    };
+
+    extraFlags = lib.mkOption {
```
aanderse, line 61: in `finix` the convention is to call this `extraArgs`

modules/services/yarr/default.nix
```diff
@@ -0,0 +1,95 @@
...
+      description = ''
+        Extra flags passed to `yarr`;
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [cfg.package];
+
+    finit.services.yarr = {
```
aanderse, line 73: you should create and assign a user and group so this service doesn't run with `root` privileges

aanderse (commented):
i think a service like `yarr` would best start off in our [community-modules](https://github.com/finix-community/community-modules) repository - would you be interested in making a PR there instead?

ar175-lol:
> i think a service like `yarr` would best start off in our [community-modules](https://github.com/finix-community/community-modules) repository - would you be interested in making a PR there instead?

I can do this tomorrow

PR #189, opened 2026-08-03
service/turnstile: init

xZecora:


aanderse:
@xZecora how does this look? are you able to test?

xZecora:
@aanderse my only concern was making it a default import, I think keeping the default import list minimal is a good idea. I changed the checks to use the `(<condition> or false)` syntax and took it out of default imports in the latest commit. Previous commit was left unchanged so if you feel like it adds unneeded complexity the you can just revert the last one. 

Otherwise it builds fine on my system.

modules/services/autologin/default.nix
r0chd, line 93: add condition for turnstiled being ready to prevent race condition

PR #192, opened 2026-08-05
modules/services/avahi: fix `extraArgs` not being used.

nostorix:


ar175-lol:
Thank you, I missed that!

PR #193, opened 2026-08-06
modules/sessiond: init

aanderse:
obviously some work to do upstream `nixpkgs`... getting `useConsoleKit` option into the `polkit` package, allowing a build truly without `systemd`...

and of course the `PAM` mess this contributes to... but let's ignore that for now, they are working on great solutions in `nixos` we will adopt as soon as they are ready :rocket:

---

ping @r0chd for review, if you don't mind :bowing_man:


r0chd:
no complaints from me

aanderse:
everything merged into `nixpkgs`, now waiting on the next `staging` cycle :tada:

willowispll:
i think it should be inside already?

modules/services/sessiond/default.nix
```diff
@@ -0,0 +1,77 @@
...
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    services.dbus.enable = true;
+    services.dbus.packages = [ cfg.package ];
+
+    services.polkit.enable = true;
+
+    finit.services.sessiond = {
```
r0chd, line 61: I fixed up the cgroup usage in `sessiond`. We should set `cgroup.delegate = true;` for the service. It won't break anything on older versions, but it will prevent breakage once the newer version is released.

modules/services/sessiond/default.nix
```diff
@@ -0,0 +1,92 @@
...
+        `sessiond` configuration. See [upstream documentation](https://tangled.org/r0chd.pl/sessiond/blob/master/docs/CONFIGURATION.md)
+        for additional details.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    services.sessiond.settings = lib.optionalAttrs config.programs.zzz.enable {
+      power = lib.mkDefault {
+        reboot = [ "/run/current-system/sw/bin/reboot" ];
```
r0chd: why gate reboot/poweroff behind `zzz` being enabled?

PR #195, opened 2026-08-07
feat: add services.nix-daemon.registry

JumpIn-Git:
This option writes /etc/nix/registry.json, i copied the logic from the nixos module. It is worth noting that it is reccomended to set
```nix
  {nixpkgs, finix, ...}: {
      nixosConfigurations.finixos = finix.lib.finixSystem {
        modules = with finix.nixosModules; [
          {
            nixpkgs.pkgs = ...;
  
            services.nix-daemon.registry.nixpkgs.to =  {
              type = "path";
              path = nixpkgs.outPath; # or `toString pkgs.path` for legacy systems or to move out of flake.nix
            };
          }
  };
```
in outputs for compatibility with legacy projects and nix repl.

This should be added to examples or set by default (using `nixpkgs.pkgs?`), although the former would fit more with finix.

I have roughly tested this like so: I build nixosConfigurations.finix.config.system.build.toplevel in a modified examples/installations/flakes/minimal with the code above, and from the `activate` file I found the related etc path, where `/nix/registry.json` is:
```json
{"flakes":[{"exact":true,"from":{"id":"nixpkgs","type":"indirect"},"to":{"path":"/nix/store/1dffznc7fgw6xnzssy1h4waxz63i11qa-source","type":"path"}}],"version":2}
```
This is my registry on my nixos system: 
```json
{"flakes":[{"exact":true,"from":{"id":"nixpkgs","type":"indirect"},"to":{"path":"/nix/store/lj0fr41wwbrx2mwq2378jg8qrns7b93d-source","type":"path"}}],"version":2}
```
Its fair to assume this will work just fine.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,11 @@
           config = lib.evalModules {
             class = "nixos";
             specialArgs = lib.recursiveUpdate { modules = self.nixosModules; } specialArgs;
-            modules = [ self.nixosModules.default ] ++ modules;
+            modules = [
+              self.nixosModules.default
+              ({pkgs,...}:{ services.nix-daemon.flake.source = toString pkgs.path; })
+            ]
+            ++ modules;
           };
         in
         config
--- a/modules/services/nix-daemon/default.nix
+++ b/modules/services/nix-daemon/default.nix
@@ -63,8 +63,13 @@ let
       ${mkKeyValuePairs (lib.filterAttrs (key: value: !(isExtra key)) cfg.settings)}
       ${mkKeyValuePairs (lib.filterAttrs (key: value: isExtra key) cfg.settings)}
     '';
+
+    flakeRefFormat = ''
+      The format of flake references is described in {manpage}`nix3-flake(1)`.
+    '';
 in
 {
+  imports = [ ./nix-flake.nix ];
   options.services.nix-daemon = {
     enable = lib.mkOption {
       type = lib.types.bool;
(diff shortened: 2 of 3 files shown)


JumpIn-Git:
let me know if the commit message should be changed or if the logic should be moved to a registry.nix!

JumpIn-Git:
alright, after rigorous testing, `toString pkgs.path` and `inputs.nixpkgs.outPath` both produce the same result (at first it wasnt, presumably because pkgs.path isnt a string) , so if the user wanted to, they could set `registry.nixpkgs = ...` in their config and not directly in the flake (using pkgs, instead of inputs) or if using legacy nix

aanderse:
thanks for contributing!

i am trying to decide which option namespace this most appropriately belongs under 🤔

is the registry a property of the daemon? i think it's purely a property of the cli, right? which might mean it is time for a `programs.nix` set of options? 

additional context: i was hoping to create a module which facilitates running a system with `nix` operating in single user mode...

what do you think? `programs.nix`, `services.nix-daemon`, or something else? 🤔

JumpIn-Git:
i also saw your nix-channel pr, maybe they can both be moved to `programs.nix`?

JumpIn-Git:
i made the flake auto set nix-daemon.flake.source, which will set the registry by default. there isnt a option for NIX_PATH yet, when this is added this can be also auto set. this would be great as it would provide compatablity with legacy projects still using `import <nixpkgs>`, and legacy nix users can use commands like `nix-shell -p hello` as nixpkgs would then be in the nix_path (also nixd wont work out of the box without this!!!)

aanderse:
sorry for the delay, i'm hoping to get back to this soon! thanks for your patience :bowing_man:

JumpIn-Git:
> sorry for the delay, i'm hoping to get back to this soon! thanks for your patience 🙇‍♂️

no problem! i think #89 should be merged with a NIX_PATH option first, as this pr kind of depends on it.

PR #198, opened 2026-08-12
modules/mdev: init

chocoblocko9:
Initialises mdev support! Notable thing that's untested is the kernel thing, no idea if it works I'm just running my own lol :P

Tested with hyprland and xorg. hyprland won't work right now thought as I'm lazy and haven't updated the modules

TODO
- [ ] Abstract the udevApi thing properly into a providers interface. Not *necessary* but it's getting crowded :smile:
- [ ] Add mdev -d support

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -25,6 +25,7 @@ let
       "gardendevd"
       "keventd"
       "mdevd"
+      "mdev"
       "seatd"
       "udev"
     ]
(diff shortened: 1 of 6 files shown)


aanderse:
btw `finit v5` has a `provides` directive which can allow us to simply use `services/devicemanager/ready` (or something like that) to wait for _any_ single device manager to be ready, instead of a specific one!

chocoblocko9:
> btw `finit v5` has a `provides` directive which can allow us to simply use `services/devicemanager/ready` (or something like that) to wait for _any_ single device manager to be ready, instead of a specific one!

oh awesome, yeah thatll clean it up a good bit

i think i got lost in the spaghetti trying to make `mdev -d` work lol, i'll get back to it soon :sob:

PR #199, opened 2026-08-13
feat(modprobe): upstream modprobe to nixos/moodules/system/boot/modprobe

s0me1newithhand7s:
image
image

repl from `fawn` -> hence eval-able 

literal upstream from nixpkgs btw


modules/boot/modprobe.nix
```diff
@@ -8,13 +8,48 @@ let
...
+        "cirrusfb"
+        "i2c_piix4"
+      ];
+      description = ''
+        Set of names of kernel modules that should not be loaded
+        automatically by the hardware probing code. This can either be
+        a list of modules or an attrset. In an attrset, names that are
+        set to `true` represent modules that will be blacklisted.
+      '';
+      apply = mods: lib.attrNames (lib.filterAttrs (_: v: v) mods);
```
aanderse: what's up with the filter function that does nothing?

aanderse (commented):
all of the changes you have made will not run until after the syslog has started in stage 2 - please confirm if that is your intention.

aside from one comment... i think this is fine, thanks for contributing 👍

modules/boot/modprobe.nix
s0me1newithhand7s: iirc that makes it possible to do both raw list and list with bool

s0me1newithhand7s:
> all of the changes you have made will not run until after the syslog has started in stage 2 - please confirm if that is your intention.

yep, i probably should stated that i aimed for stage 2, rather then initrd / stage 1 with modprobe upstream

aanderse:
i pushed some changes... what do you think? anyone brave enough to test this out? :sweat_smile:

modules/programs/modprobe/default.nix
```diff
@@ -0,0 +1,109 @@
...
+let
+  cfg = config.programs.modprobe;
+in
+{
+  imports = [
+    (lib.mkRenamedOptionModule
+      [ "boot" "blacklistedKernelModules" ]
+      [ "programs" "modprobe" "blacklist" ]
+    )
+    (lib.mkRenamedOptionModule [ "boot" "extraModprobeConfig" ] [ "programs" "modprobe" "extraConfig" ])
```
s0me1newithhand7s: my idea was to upstream `modprobe` bc of convenience, so to put in `boot`
if your vision is `progrmas.modprobe` - that's okay!

modules/programs/modprobe/default.nix
```diff
@@ -0,0 +1,109 @@
...
+        {manpage}`modprobe.d(5)` for details.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.etc."modprobe.d/ubuntu.conf".source = "${pkgs.kmod-blacklist-ubuntu}/modprobe.conf";
+    environment.etc."modprobe.d/debian.conf".source = pkgs.kmod-debian-aliases;
+
+    environment.etc."modprobe.d/00-nixos.conf".text = ''
```
s0me1newithhand7s: and maybe change 00-nixos-conf to 00-finix-conf?

xZecora:
There are a few other things in the services directory that use `kmod`, specifically `docker`, `system76-scheduler`, and `incus`. Do they need to use the new option as well or is there a reason for them to keep `kmod`?

aanderse:
> There are a few other things in the services directory that use `kmod`, specifically `docker`, `system76-scheduler`, and `incus`. Do they need to use the new option as well or is there a reason for them to keep `kmod`?

i did not look at their source code and thought it might not be a great idea to change these specific services without some sort of testing or verification

PR #204, opened 2026-08-15
virtualisation: add android module (binder/ashmem + waydroid)

y0usaf:
Loads `binder_linux` and `ashmem_linux` kernel modules and ships the Waydroid dbus policy (`id.waydro.Container`) so Waydroid can boot an Android container.

Usage:
```nix
virtualisation.android.enable = true;
```

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -73,3 +73,8 @@ in
 }
 // programModules
 // serviceModules
+// {
+  # virtualisation
+
+  android = ./virtualisation/android.nix;
+}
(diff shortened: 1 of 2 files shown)


aanderse:
looks good! should we consider making this an optional module, though?

y0usaf:
We totally should! opt-in via `nixosModules.virtualisation.android`:
```nix
imports = [ finix.nixosModules.virtualisation.android ];
virtualisation.android.enable = true;
```

PR #205, opened 2026-08-16
programs/coreutils: added options for a handful of utils

xZecora:
Opening this as a draft just so other people can see and add on any work.

### DIFF
--- a/modules/filesystems/default.nix
+++ b/modules/filesystems/default.nix
@@ -140,8 +140,8 @@ let
                 ${lib.optionalString re.allowDiscards "--allow-discards"} \
                 -d ${lib.escapeShellArg re.source} \
                 ${lib.escapeShellArg sw.device} ${lib.escapeShellArg name}
-              ${pkgs.util-linuxMinimal}/bin/mkswap /dev/mapper/${name}
-              ${pkgs.util-linuxMinimal}/bin/swapon -o ${lib.escapeShellArg (lib.concatStringsSep "," options)} /dev/mapper/${name}
+              ${config.programs.utilLinux.package}/bin/mkswap /dev/mapper/${name}
+              ${config.programs.utilLinux.package}/bin/swapon -o ${lib.escapeShellArg (lib.concatStringsSep "," options)} /dev/mapper/${name}
             ''
           );
         };
--- a/modules/programs/coreutils/default.nix
+++ b/modules/programs/coreutils/default.nix
@@ -5,6 +5,7 @@
   ...
 }:
 {
+  imports = [ ./internal-utils.nix ];
   options.programs.coreutils.package = lib.mkOption {
     type = lib.types.package;
     default = pkgs.coreutils;
--- a/modules/programs/coreutils/internal-utils.nix
+++ b/modules/programs/coreutils/internal-utils.nix
@@ -1,0 +1,34 @@
+{ lib, pkgs, ... }:
+{
+  options = {
+    programs.sed.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.gnused;
+    };
+
+    programs.grep.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.gnugrep;
+    };
+
+    programs.tar.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.gnutar;
+    };
+
+    programs.which.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.which;
+    };
+
+    programs.utilLinux.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.util-linux;
+    };
+
+    programs.findUtils.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.findutils;
+    };
+  };
+}
--- a/modules/security/wrappers/default.nix
+++ b/modules/security/wrappers/default.nix
@@ -255,8 +255,8 @@ in
       in
       {
         # These are mount related wrappers that require the +s permission.
-        mount = mkSetuidRoot "${lib.getBin pkgs.util-linux}/bin/mount";
-        umount = mkSetuidRoot "${lib.getBin pkgs.util-linux}/bin/umount";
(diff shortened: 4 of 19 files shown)


modules/security/wrappers/default.nix
```diff
@@ -255,8 +255,8 @@ in
       in
       {
         # These are mount related wrappers that require the +s permission.
-        mount = mkSetuidRoot "${lib.getBin pkgs.util-linux}/bin/mount";
-        umount = mkSetuidRoot "${lib.getBin pkgs.util-linux}/bin/umount";
+        mount = mkSetuidRoot "${config.programs.utilLinux.package}/bin/mount";
```
s0me1newithhand7s, line 258: what if `${lib.getExe' config.programs.utilLinux.package "mount"}`?

modules/services/fstrim/default.nix
```diff
@@ -40,7 +40,7 @@ in
         fstrim = {
           inherit (cfg) interval;
 
-          command = "${pkgs.util-linux}/bin/fstrim --listed-in /etc/fstab:/proc/self/mountinfo --verbose --quiet-unsupported";
+          command = "${config.programs.utilLinux.package}/bin/fstrim --listed-in /etc/fstab:/proc/self/mountinfo --verbose --quiet-unsupported";
```
s0me1newithhand7s: what if `${lib.getExe' config.programs.utilLinux.package "fstrim"}`?

modules/filesystems/default.nix
```diff
@@ -140,8 +140,8 @@ let
                 ${lib.optionalString re.allowDiscards "--allow-discards"} \
                 -d ${lib.escapeShellArg re.source} \
                 ${lib.escapeShellArg sw.device} ${lib.escapeShellArg name}
-              ${pkgs.util-linuxMinimal}/bin/mkswap /dev/mapper/${name}
-              ${pkgs.util-linuxMinimal}/bin/swapon -o ${lib.escapeShellArg (lib.concatStringsSep "," options)} /dev/mapper/${name}
+              ${config.programs.utilLinux.package}/bin/mkswap /dev/mapper/${name}
```
s0me1newithhand7s: `${lib.getExe' config.programs.utilLinux.package "mkswap"}`

modules/services/gardendevd/default.nix
```diff
@@ -140,8 +140,8 @@ in
             substituteInPlace "$i" \
               --replace-quiet \"/sbin/modprobe \"${pkgs.kmod}/bin/modprobe \
               --replace-quiet \"/sbin/mdadm \"${pkgs.mdadm}/sbin/mdadm \
-              --replace-quiet \"/sbin/blkid \"${pkgs.util-linux}/sbin/blkid \
-              --replace-quiet \"/bin/mount \"${pkgs.util-linux}/bin/mount \
+              --replace-quiet \"/sbin/blkid \"${config.programs.utilLinux.package}/sbin/blkid \
```
s0me1newithhand7s: `${lib.getExe' config.programs.utilLinux.package "blkid"}` ?

s0me1newithhand7s:
i'm not offending but `lib.getExe'` is cool 👀

modules/programs/coreutils/internal-utils.nix
```diff
@@ -0,0 +1,34 @@
...
+    programs.which.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.which;
+    };
+
+    programs.utilLinux.package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.util-linux;
+    };
+
```
aanderse, line 28:
```suggestion
```

there isn't a true alternative `util-linux` package, so i think we should drop this one... also, we hard depend on features from this package in a number of places, so there isn't _really_ an option here at the moment

modules/programs/coreutils/internal-utils.nix
```diff
@@ -0,0 +1,34 @@
...
+    programs.findUtils.package = lib.mkOption {
```
aanderse, line 29:
```suggestion
    programs.find.package = lib.mkOption {
```

🤔

modules/services/incus/default.nix
```diff
@@ -60,14 +60,14 @@ in
           btrfs-progs
           cdrkit
           config.programs.coreutils.package
+          config.programs.grep.package
+          config.programs.sed.package
+          config.programs.tar.package
+          config.programs.findUtils.package
```
aanderse, line 66: i doubt this would work - if the user specifies `busybox`, as an example, we don't know that `incus` source code isn't using incompatible options

having a minimal profile where it is compatible is cool... but if `incus` *wants* `gnused` it should use `gnused`

a future project could be to see if `incus` is compatible with other implementations, but let's defer that for now

modules/services/system76-scheduler/default.nix
```diff
@@ -58,8 +58,8 @@ in
 
       # TODO: now we're hijacking `env` and no one else can use it...
       path = with pkgs; [
-        kmod
-        gnutar
+        pkgs.kmod
+        config.programs.tar.package
```
aanderse, line 62: do we know `busybox` or friends would suffice here?

modules/programs/coreutils/default.nix
```diff
@@ -5,6 +5,7 @@
   ...
 }:
 {
+  imports = [ ./internal-utils.nix ];
```
aanderse, line 8: i think an additional file doesn't benefit us much here - please inline the code directly

modules/programs/coreutils/internal-utils.nix
```diff
@@ -0,0 +1,34 @@
+{ lib, pkgs, ... }:
+{
+  options = {
```
aanderse, line 3: do we want or need `enable` options for any of these? is there _ever_ a case where you don't want `which` added to the system path, for example? 🤔

i'm considering extremely constrained embedded devices or images which aren't ever meant to have a shell...

obviously `coreutils` is _always_ needed... but maybe there are some here which are not?

aanderse (changes requested):
i like what you're doing here 👍

PR #212, opened 2026-08-20
sddm: options to change package and add extraPackages

dmkhitaryan:
This PR adds the option for users to pass `extraPackages` into the `sddm` wrapper instead of adding those globally to system packages. As an example, this allows users to pass libraries needed for certain themes, such as `qtmultimedia`.

### DIFF
--- a/modules/services/sddm/default.nix
+++ b/modules/services/sddm/default.nix
@@ -10,6 +10,10 @@ let
 
   format = pkgs.formats.ini { };
   configFile = format.generate "sddm.conf" cfg.settings;
+
+  package' = cfg.package.override (prev: {
+    extraPackages = prev.extraPackages or [ ] ++ cfg.extraPackages;
+  });
 in
 {
   imports = [ modules.xorg ];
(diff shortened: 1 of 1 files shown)


modules/services/sddm/default.nix
```diff
@@ -23,6 +27,33 @@ in
...
     };
 
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.kdePackages.sddm;
+      defaultText = lib.literalExpression "pkgs.kdePackages.sddm";
+      description = "The sddm package to use.";
+    };
+
+    extraPackages = lib.mkOption {
```
aanderse, line 39: is there value in us adding our own `extraPackages` wrapper, or should we simply explain in the `package` options `example`, like

```nix
with pkgs.kdePackages; sddm.override {
  extraPackages = [
    breeze-icons
    kirigami
    libplasma
    plasma5support
    qtsvg
    qtvirtualkeyboard
    qtmultimedia
  ];
};
```

aanderse (commented):
thank you so much for looking into this! i left a comment - what do you think?

modules/services/sddm/default.nix
dmkhitaryan: Not a strictly definitive value; I just think having `extraPackages` as an option leads to a more friendly experience.

aanderse: ok 👍

i would like to put a standard set of options together that have a well defined meaning for each module that implements them... `extraPackages` can fit into this

thanks!

aanderse:
i renamed the `sddm` variable to `package'` to make scanning the code a bit easier on my eyes... so much `sddm` text already in the file :sweat_smile:

thanks so much for the PR!

PR #219, opened 2026-08-24
add a commonPackages namespace for modules to pass around dependencies

xZecora:


xZecora:
`gardendev` and `xorg` have an example usage of my idea for the `global.libudev` option so I think thats the most interesting change, everything else is a basic use case. There's a few comments to explain some choices here and there but if anything is confusing please ask for clarification.

Another namespace may be ideal just because of sheer length, I added shorthand to the top of basically every single file I touched which isn't the best. I was also somewhat liberal with the use of setting everything globally. In the long run that might not be ideal.

modules/finit/default.nix
```diff
@@ -5,6 +5,8 @@
   ...
 }:
 let
+  globalPackages = config.environment.commonPackages.global;
+  finitPackages = config.environment.commonPackages.global;
```
dmkhitaryan, line 9:
```suggestion
  finitPackages = config.environment.commonPackages.finit;
```
I think it's also worth keeping in mind with this level of module writing, it might be much easier to make mistakes.

dmkhitaryan:
> Another namespace may be ideal just because of sheer length, I added shorthand to the top of basically every single file I touched which isn't the best. I was also somewhat liberal with the use of setting everything globally. In the long run that might not be ideal.

In fairness, since we are pulling through `config`, it's going to be global no matter what. I don't think we can do without that.

PR #223, opened 2026-08-28
modules/cosmic-session: init

r0chd:
this change initializes `cosmic-session` and minimal set of programs needed to run it. This change doesn't include optional dependencies that can be started by `cosmic-session` like `cosmic-launcher`, `cosmic-idle` or icon/sound themes


aanderse (approved):
awesome! thank you!

PR #225, opened 2026-08-28
providers.privileges (doas): add missing 'args'

dmkhitaryan:
Privileges setup for `doas` misses `args` flag, meaning that rules written in the likes of:
```Nix
providers.privileges.rules = [
{
  users = [ "user" ];
  groups = [ ];
  runAs = "root";
  requirePassword = false;
  command = "/run/current-system/sw/bin/initctl";
  args = [ "suspend" ];
}
```
Will fail to build.

### DIFF
--- a/modules/programs/doas/providers.privileges.nix
+++ b/modules/programs/doas/providers.privileges.nix
@@ -25,10 +25,10 @@
           in
           ''
             ${lib.concatMapStringsSep "\n" (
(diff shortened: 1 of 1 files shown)


aanderse:
@xZecora look good to you? i'm not a `doas` guy...

xZecora:
Looks good to me, just adds syntax that was missing for some reason.

xZecora:
Author confirmed working as intended earlier but I just confirmed it myself.

aanderse (approved):
thank you both!

PR #226, opened 2026-08-29
modules/getty: do not enable by default

aanderse:
i did not test yet... :eyes:

### DIFF
--- a/modules/finit/default.nix
+++ b/modules/finit/default.nix
@@ -22,6 +22,11 @@ in
         assertion = lib.versionAtLeast cfg.package.version "4.16";
         message = "finit version must be at least 4.16";
       }
+
+      {
+        assertion = config.finit.ttys != { };
+        message = "you have not defined any ttys; consider importing and enabling the getty module";
+      }
     ];
 
     # TODO: decide a reasonable default here... user can override if needed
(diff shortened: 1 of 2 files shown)


xZecora:
I'm not very familiar with truly headless systems that are purely `ssh` access, but maybe we should have an option that lets users disable the warning? Just an acknowledgement that "Yes I know this is risky but I don't want to see the warning".

Sorry if this is a basically non-existent use case but I just know I hate seeing warnings that I know I can ignore.

aanderse:
@willowispll suggested the assertion as a temporary measure so we don't catch any existing users off guard

the idea is to remove the assertion in a month or so because your scenario is relevant 👍

willowispll:
tested, works great!

modules/finit/default.nix
```diff
@@ -22,6 +22,11 @@ in
         assertion = lib.versionAtLeast cfg.package.version "4.16";
         message = "finit version must be at least 4.16";
       }
+
+      {
+        assertion = config.finit.ttys != { };
```
willowispll, line 27: this lets to build without having getty enabled if u have ly

willowispll (approved):
LGTM!

modules/finit/default.nix
aanderse: an interesting game of roulette no doubt 😉

PR #227, opened 2026-08-29
modules/environment: set default /bin/sh to dash

aanderse:


xZecora (commented):
Rebuilt using this and commented out my line doing the same thing. My system didn't even generate a new `/nix/store/...-finix-system` so I'm inclined to believe it should be fine, at least as far as my closure. Mileage may vary with different modules/userspaces.

Maybe we should put up a warning for a few weeks reminding people the default `/bin/sh` is no longer bash, just in case they use it in personal scripts or the like.

aanderse:
YOLO

PR #228, opened 2026-08-29
modules/environment: drop some of the less controversial packages fro…

aanderse:
…m system closure

### DIFF
--- a/modules/environment/path/default.nix
+++ b/modules/environment/path/default.nix
@@ -34,33 +34,21 @@
     environment.systemPackages = with pkgs; [
       acl
       attr
-      bzip2
-      cpio
-      curl
-      diffutils
       findutils
       getent
       getconf
-      gzip
-      xz
       less
       libcap
       ncurses
-      netcat
-      mkpasswd
       procps
       su
       time
       util-linux
       which
-      zstd
 
       bashInteractive
       gawk
       gnugrep
-      gnupatch
-      gnused
-      gnutar
     ];
 
     environment.pathsToLink = [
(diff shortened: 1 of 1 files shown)


modules/environment/path/default.nix
```diff
@@ -72,15 +60,9 @@
...
-      # TODO: trim this list down
-      "/share/emacs"
-      "/share/hunspell"
-      "/share/org"
+      # TODO: trim this list down?
       "/share/themes"
       "/share/vulkan"
-      "/share/kservices5"
-      "/share/kservicetypes5"
-      "/share/kxmlgui5"
```
aanderse, line 83: @deathbymanatee is this ok? does this cause problems for `lxqt`? if so, can we move these definitions into that module?

modules/environment/path/default.nix
```diff
@@ -72,15 +60,9 @@
...
       "/share/thumbnailers"
       "/share/wayland-sessions"
```
willowispll, line 67: can we also make this optional?

modules/environment/path/default.nix
```diff
@@ -72,15 +60,9 @@
       "/lib" # FIXME: remove and update debug-info.nix
       "/sbin"
 
-      # TODO: trim this list down
-      "/share/emacs"
-      "/share/hunspell"
-      "/share/org"
+      # TODO: trim this list down?
       "/share/themes"
       "/share/vulkan"
```
willowispll: and this can be bundled with finit graphical level?

modules/environment/path/default.nix
willowispll: the `"/share/wayland-sessions"`

modules/environment/path/default.nix
willowispll: `"/share/vulkan"`

modules/environment/path/default.nix
aanderse: i think most things here can be optional/only added when needed...

it would take some work to remove this line... every wayland based module would need to add it

that sounds like a full PR right there... one i would happily merge

PR #229, opened 2026-08-29
remove not needed packages from modules/environment/path

willowispll:


modules/environment/path/default.nix
```diff
@@ -36,18 +36,12 @@
       attr
       findutils
       getent
-      getconf
-      less
       libcap
       ncurses
       procps
-      su
```
willowispll, line 44: this is pulled from shadow (enabled+imported automatically)

modules/environment/path/default.nix
```diff
@@ -36,18 +36,12 @@
       attr
       findutils
       getent
-      getconf
```
willowispll: never referenced

modules/environment/path/default.nix
```diff
@@ -36,18 +36,12 @@
...
-      less
       libcap
       ncurses
       procps
-      su
-      time
       util-linux
-      which
```
willowispll: why?

modules/environment/path/default.nix
```diff
@@ -36,18 +36,12 @@
...
 
       bashInteractive
-      gawk
```
willowispll: also never referenced

modules/environment/path/default.nix
```diff
@@ -32,22 +32,11 @@
...
-      acl
-      attr
       findutils
-      getent
-      getconf
-      less
-      libcap
       ncurses
-      procps
-      su
```
aanderse, line 44: should not _remove_ this, you should add it to the `shadow` module under `users.users.root.packages` - if not then `root` will end up using the `setuid` binary, which isn't what we want... even worse, if a user disables the `setuid` binary then `root` will have no `su` at all, despite `shadow` being enabled... this would be incorrect behaviour

modules/environment/path/default.nix
```diff
@@ -32,22 +32,11 @@
 
   config = {
     environment.systemPackages = with pkgs; [
-      acl
-      attr
       findutils
-      getent
-      getconf
-      less
```
aanderse, line 40: hmm... without any replacement i feel like a number of tools won't work correctly - many tools assume there is a pager avaliable...

modules/environment/path/default.nix
```diff
@@ -32,22 +32,11 @@
 
   config = {
     environment.systemPackages = with pkgs; [
-      acl
-      attr
       findutils
-      getent
```
aanderse, line 38: good catch! this likely belongs in a future `nsswitch` module!

modules/environment/path/default.nix
```diff
@@ -32,22 +32,11 @@
 
   config = {
     environment.systemPackages = with pkgs; [
-      acl
```
aanderse, line 35: i have a bad feeling about removing this... also, `acl` is absolutely in your system closure already, i do not believe you can create a nix based system without it effectively? or maybe i'm mistaken?

modules/environment/path/default.nix
```diff
@@ -32,22 +32,11 @@
...
-      attr
       findutils
-      getent
-      getconf
-      less
-      libcap
       ncurses
-      procps
```
aanderse, line 43: i'm not comfortable removing this specifically tonight. i would need more time to think about this... i suggest removing that in a future PR


```
nixpkgs on  master [$] on ☁️
❯ ls ./result/bin
free*  hugetop*  kill*  pgrep*  pidof*  pidwait*  pkill*  pmap*  ps*  pwdx*  slabtop*  sysctl*  tload*  top*  uptime*  vmstat*  w*  watch*
```

🫠

aanderse (changes requested):

modules/environment/path/default.nix
willowispll: su in the shadow package already, dont need to add anything?

modules/environment/path/default.nix
willowispll: can u give me any example to test?

modules/environment/path/default.nix
willowispll: i have only 1 acl - from steam:>

modules/environment/path/default.nix
willowispll: procps stays from finit, but why user wants to use it? 
idk about less tho, its in my busybox 
libcap is in closure from suid (and steam?), why we need it here?

willowispll:
i dont see here anything actually breaking the system (maybe less?)

aanderse:
i can understand disregarding my comments that start with "i have a bad feeling" :sweat_smile: ... but i did point out that at least the `su` change is _incorrect_ and should be done differently, and the `less` removal likely _will_ break a number of systems... so we might need to give that further consideration and research

willowispll:
yes, i understand, lets put that to more testing at least (i still think we need to remove most of the stuff from environment/path, its either not used or should not have been here at the first place. like that lib, why its in a top-level, its not even an executable? and most of the stuff in nixpkgs is self-containered (idk if i can call it like that) with it's deps)

i was not expecting this to go quickly, so all good

z4il-i:
```
    environment.systemPackages = with pkgs; [
      bzip2
      cpio
      ncurses
      util-linux
      su
      zstd
    ];
```
This is the list I have and my system functions as expected, dmesg showed no errors nor warnings, haven't had any issues thus far

PR #230, opened 2026-08-30
modules/nix-garbage-collect: init

z4il-i:
Adds the service nix-garbage-collect meant for unifying any garbage collection service into one location.
Adds 3 basic backends, `nix store`, `nix-store`, `nix-garbage-collect`


aanderse (changes requested):
as mentioned in the `nh` PR i think it would be best to offer a concrete implementation now instead of an abstraction - would you be interested in changing this into a `nix-collect-garbage` service that is loosely based on the `nix.gc` module?

z4il-i:
I think it might be better to then to make a loosely based ``nix.gc`` (and remove this) under ``services.nix-daemon.gc``

aanderse:
what if someone runs a daemonless nix? i would like to support this use case.

z4il-i:
Just make the option not dependent on the nix daemon being active, but it still lives under there, which sounds the most reasonable to me as after all, it is a nix base service, a daemon if you will, for auto cleaning nix store with nix-garbage-collect

aanderse:
i am not sure i follow... all (well... many) commands go through the daemon if it is running ... but otherwise do not

are you familiar with operating nix without a daemon? i feel like most people aren't and putting this service under this namespace doesn't accurately reflect the situation...

z4il-i:
I am not familiar with it no, just thought it would make more sense to be under it but now i see it won't
I will make the change tomorrow

PR #231, opened 2026-08-30
modules/nh: init

z4il-i:
Adds the NH module which differs from the one in nixpkgs by removing the dependency on `systemd.timers` for the `nh.clean` option seat, instead using the new `services.nix-garbage-collect` module in lue of the `nh.clean` option set, it also adds 2 new options, `nh.file` and `nh.attrp` which like `nh.flake` are just shortcuts to their respective environment variables

### DIFF
--- a/modules/programs/nh/default.nix
+++ b/modules/programs/nh/default.nix
@@ -1,0 +1,89 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.programs.nh;
+in
+{
+  options = {
+    services.nix-garbage-collect.backend = lib.mkOption {
+      type = lib.types.enum [ "nh" ];
+    };
+    programs.nh = {
+      enable = lib.mkEnableOption "nh, yet another Nix CLI helper";
+
+      package = lib.mkPackageOption pkgs "nh" { };
+
+      flake = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          The string that will be used for the `NH_FLAKE` environment variable.
+
+          `NH_FLAKE` is used by nh as the default flake for performing actions, such as
+          `nh os switch`. This behaviour can be overriden per-command with environment
+          variables that will take priority.
+
+          - `NH_OS_FLAKE`: will take priority for `nh os` commands.
+          - `NH_HOME_FLAKE`: will take priority for `nh home` commands.
+          - `NH_DARWIN_FLAKE`: will take priority for `nh darwin` commands.
+
+          The formerly valid `FLAKE` is now deprecated by nh, and will cause hard errors
+          in future releases if `NH_FLAKE` is not set.
+
+          `NH_FLAKE` can point to either a folder containing a flake, or to an outside repository containing the flake.
+        '';
+      };
+      file = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          The string that will be used for the `NH_FILE` environment variable.
+
+          `NH_FILE` is used by nh as the default configuration file for performing actions, such as
+          `nh os switch`. This behaviour can be overriden per-command with environment variables
+          that will take priority
+        '';
+      };
+      attrp = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = ''
+          The string that will be used for the `NH_ATTRP` environment variable.
+
+          `NH_ATTRP` is used by nh as the default attribute for performing actions, such as
+          `nh os switch`. This behaviour can be overriden per-command with environment variables
+          that will take priority
+        '';
+      };
+    };
+  };
+  config = {
+    assertions = [
+      {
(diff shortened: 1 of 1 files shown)


modules/programs/nh/default.nix
```diff
@@ -0,0 +1,89 @@
...
+  ...
+}:
+let
+  cfg = config.programs.nh;
+in
+{
+  options = {
+    services.nix-garbage-collect.backend = lib.mkOption {
+      type = lib.types.enum [ "nh" ];
+    };
```
aanderse, line 14: i have decided that an abstraction like this would cause an inconsistency with all other modules which serve the same purpose so needs a higher level review... can you please drop the backend stuff for now and keep this as a simple single `nh` module?

modules/programs/nh/default.nix
```diff
@@ -0,0 +1,89 @@
...
+    programs.nh = {
+      enable = lib.mkEnableOption "nh, yet another Nix CLI helper";
```
aanderse, line 16: please follow the `enable` pattern that all other `programs` modules follow

modules/programs/nh/default.nix
```diff
@@ -0,0 +1,89 @@
...
+
+      package = lib.mkPackageOption pkgs "nh" { };
```
aanderse, line 18: please follow the `package` pattern that all other `programs` modules follow

modules/programs/nh/default.nix
```diff
@@ -0,0 +1,89 @@
...
+
+      flake = lib.mkOption {
```
aanderse, line 20: i wonder if it is worth changing this option into a `settings` with a `freeformType`, where we explicitly declare `NH_FLAKE`, `NH_FILE`, and `NH_ATTRP` as nullable options... 🤔

that feels like the least opinionated thing to do here

modules/programs/nh/default.nix
```diff
@@ -0,0 +1,89 @@
...
+      variables = lib.mkMerge [
+        (lib.mkIf (cfg.flake != null) { NH_FLAKE = cfg.flake; })
+        (lib.mkIf (cfg.file != null) { NH_FILE = cfg.file; })
+        (lib.mkIf (cfg.attrp != null) { NH_ATTRP = cfg.attrp; })
+      ];
+    };
+
+    services.nix-garbage-collect = lib.mkIf (config.services.nix-garbage-collect.backend == "nh") {
+      command = "${lib.getExe cfg.package} clean all ${config.services.nix-garbage-collect.extraArgs}";
+    };
```
aanderse, line 87: see comment above about dropping abstraction - this can become `providers.schedulers.nh` or some such

modules/programs/nh/default.nix
aanderse, line 20: i would like to validate whether remote flake references work here if you don't mind... very powerful if they do!

aanderse (changes requested):

modules/programs/nh/default.nix
z4il-i: Should I revert to https://github.com/finix-community/community-modules/blob/main/modules/programs/nh/default.nix#L83-L86 or remove the clean option entirely?

modules/programs/nh/default.nix
z4il-i: Alright, but may I ask why?

modules/programs/nh/default.nix
z4il-i: I like the idea, better than what I and nixpkgs did!

modules/programs/nh/default.nix
aanderse: it renders nicely! and consistency.

image

includes a link to upstream for better discoverability. i only wish `nixos` would adopt this!

modules/programs/nh/default.nix
aanderse: i didn't think we need a provider - providers aren't for end users really

having a clean option is fine, i think i would use it

modules/programs/nh/default.nix
z4il-i: Alright, it will be done :D

modules/programs/nh/default.nix
z4il-i: Do I then return  the ``nh.clean`` option or remove it entirely?

aanderse: oh boy... ignore that comment

please add `.clean.*` options back but don't add a new provider

sorry for the noise

modules/programs/nh/default.nix
```diff
@@ -0,0 +1,276 @@
...
+          !(
+            (
+              cfg.settings.NH_FLAKE != null
+              || cfg.settings.NH_OS_FLAKE != null
+              || cfg.settings.NH_HOME_FLAKE != null
+              || cfg.settings.NH_DARWIN_FLAKE != null
+            )
+            && cfg.settings.NH_FILE != null
+          );
+        message = "NH_FILE and may FLAKE option can not be set at the same time, they are opposite components";
```
es-sai-fi, line 241:
```suggestion
        message = "NH_FILE and FLAKE options are mutually exclusive";
```

PR #232, opened 2026-08-30
tree-wide: enable sysklogd in modules with reliance on it

dmkhitaryan:
Inspired by the not-so-pleasant situation where a display manager like `ly` will never launch because it expects `syslogd` to be ready. This will never occur since a user has to manually enable `services.sysklogd`. This PR adds enabling of `sysklogd` to every module that relies on it.

### DIFF
--- a/modules/hardware/console.nix
+++ b/modules/hardware/console.nix
@@ -64,6 +64,7 @@ in
   };
 
   config = lib.mkIf cfg.enable {
+    services.sysklogd.enable = true;
     environment.systemPackages = [ pkgs.kbd ];
 
     # Include binary keymap in the initramfs.
--- a/modules/programs/ifupdown-ng/default.nix
+++ b/modules/programs/ifupdown-ng/default.nix
@@ -161,6 +161,7 @@ in
     programs.ifupdown-ng.extraArgs = [ "-a" ] ++ lib.optionals cfg.debug [ "-v" ];
 
     environment.systemPackages = [ cfg.package ];
+    services.sysklogd.enable = true;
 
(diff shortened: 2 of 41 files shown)


aanderse:
but what if the user is running [rsyslog](https://github.com/finix-community/finix/blob/4da4923c1b973c48dbf3fddfe19beefbe4f22955/modules/services/rsyslog/default.nix#L51) instead...

dmkhitaryan:
> but what if the user is running [rsyslog](https://github.com/finix-community/finix/blob/4da4923c1b973c48dbf3fddfe19beefbe4f22955/modules/services/rsyslog/default.nix#L51) instead...

Good point, that's the other side of the problem. But leaving new users stuck because they don't they need to enable syslogd provider themselves is not great either.

dmkhitaryan:
Then the best case is to add assertions tree-wide @aanderse. What do you say?
Something similar to the provider scheduler:
```
evaluation warning: no syslogd service has been enabled, but the following modules rely on it conditionally:
                    ly
                    ...
Enable a syslogd service to ensure aforementioned modules work correctly.
```

aanderse:
again... basically _every_ service depends on a `syslog`, either directly (`conditions = [ "service/syslogd/ready" ];`), or indirectly (`conditions = [ "service/dbus/ready" ];`, where `dbus` depends on `syslogd`, for example)

there is no point in listing what depends... it will be _every_ service

dmkhitaryan:
Closing because given what we want to do (not write assertion to every file, while also not enabling this service by default). At that point it comes down to a documentation or pointing people to profiles in community modules.

PR #236, opened 2026-08-31
feat(wayland): initialize Umbriel compositor option

s0me1newithhand7s:
since my PR in umbriel directly was rejected - i did this :D 
since `pkgs.umrbiel` essentially is `wlroots` and `scenefx` i don't think it would be wildly different from `pkgs.sway`, but further test are needed

### DIFF
--- a/modules/programs/umbriel/default.nix
+++ b/modules/programs/umbriel/default.nix
@@ -1,0 +1,72 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  cfg = config.programs.umbriel;
+
+  sessionFile = pkgs.writeTextDir "share/wayland-sessions/umbriel.desktop" ''
+    [Desktop Entry]
+    Comment=Umbrel, a Wayland compositor built on wlroots and SceneFX
+    DesktopNames=umbriel
+    Exec=${lib.getExe' pkgs.dbus "dbus-run-session"} -- ${lib.getExe' cfg.package "start-umbriel"}
+    Name=Umbriel
+    Type=Application
+  '';
+
+  # gardendevd needs libudev-garden; mdevd/keventd need libudev-zero
+  udevApi =
+    if config.services.gardendevd.enable then
+      pkgs.libudev-garden
+    else if config.services.mdevd.enable || config.services.keventd.enable then
+      pkgs.libudev-zero
+    else
+      null;
+
+  libinput = pkgs.libinput.override (
+    lib.optionalAttrs (udevApi != null) {
+      udev = udevApi;
+      wacomSupport = false;
+    }
+  );
+in
+{
+  options.programs.umbriel = {
+    enable = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable [umbriel](${pkgs.umbriel.meta.homepage}).
+      '';
+    };
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.umbriel.override (
+        o:
+        let
+          wlrootsAttrs = lib.head (lib.filter (lib.hasPrefix "wlroots") (lib.attrNames o));
+        in
+        {
+          inherit libinput;
+          ${wlrootsAttrs} = o.${wlrootsAttrs}.override { inherit libinput; };
+        }
+      );
+      defaultText = lib.literalExpression "pkgs.umbriel";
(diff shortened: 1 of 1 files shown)


modules/programs/umbriel/default.nix
```diff
@@ -0,0 +1,65 @@
...
+  udevApi =
+    if config.services.gardendevd.enable then
+      pkgs.libudev-garden
+    else if config.services.mdevd.enable || config.services.keventd.enable then
+      pkgs.libudev-zero
+    else
+      null;
+
+  overrideAttrs = lib.optionalAttrs (udevApi != null) {
+    eudev = udevApi;
```
aanderse: i saw a `wlroots_0_20` argument (see `labwc` module) in the `nixpkgs` package, but no `eudev` argument, which means this won't evaluate if using a device manager other than `udev`

aanderse (changes requested):
very cool! thanks for contributing! are you able to test or find someone who can?

s0me1newithhand7s:
> very cool! thanks for contributing! are you able to test or find someone who can?

test? yep

modules/programs/umbriel/default.nix
s0me1newithhand7s: oof, that's sound like a thing to dive in

s0me1newithhand7s:
image
non-evalutable error! glad im doing repl tests before real roll-out

s0me1newithhand7s:
image
some force pushes after

modules/programs/umbriel/default.nix
```diff
@@ -0,0 +1,72 @@
...
+      null;
+
+  libinput = pkgs.libinput.override (
+    lib.optionalAttrs (udevApi != null) {
+      udev = udevApi;
+      wacomSupport = false;
+    }
+  );
+
+  wlroots_0_20 = pkgs.wlroots_0_20.override {
```
aanderse: this will break when they bump to `wlroots_0_21` - and updating it will cause people who haven't updated `nixpkgs` to break as well

see `labwc` how to properly handle this

aanderse (changes requested):

modules/programs/umbriel/default.nix
s0me1newithhand7s: np!

modules/programs/umbriel/default.nix
```diff
@@ -0,0 +1,71 @@
...
+      type = lib.types.package;
+      default = pkgs.umbriel.override {
+        inherit libinput;
+
+        wlroots_0_20 = pkgs.wlroots_0_20.override {
+          inherit libinput;
+          # since xwayland done via xwayland-sattelite sway part
+          # of xwayland is useless
+        };
+      };
```
aanderse:
```suggestion
      default = pkgs.umbriel.override (
        o:
        let
          wlrootsAttr = lib.head (lib.filter (lib.hasPrefix "wlroots") (lib.attrNames o));
        in
        {
          inherit libinput;
          ${wlrootsAttr} = o.${wlrootsAttr}.override { inherit libinput; };
        }
      );
```

can you please test this?

s0me1newithhand7s: in `umbriel`? probably yes
but i feel like this better to be atm tested at stable already working compositor, which i'll do asap

aanderse: this pattern is already used in `labwc`, `regreet`, etc...

s0me1newithhand7s: if it working there then np

s0me1newithhand7s: image
oh maybe this is why i dunno, bc my local copy is this old

s0me1newithhand7s:
so it's silently hangs on my system
ig this is another issue to be open in `umbriel`

Ag1O1:
The module seems to work on my system, I'm not getting any hanging. At first it kept crashing but then i found out that i forgot to add my user to the render group. Now its working fine.

s0me1newithhand7s:
> The module seems to work on my system, I'm not getting any hanging. At first it kept crashing but then i found out that i forgot to add my user to the render group. Now its working fine.

im in all requered groups - `sway`, `mango` and `niri` work fine with module (hence pacthed package)

s0me1newithhand7s:
ig our job is done here, it runs under `finix`, sadly, under nested, but runs

PR #237, opened 2026-09-01
modules/tmpfiles: add finer control on tmpfiles auto clean

deathbymanatee:


deathbymanatee:
a positive side effect is that it removes the warning message on rebuilds if you don't have a cron service active (which may or may not be the reason i did this)

aanderse:
thank you! any chance you can enable this on the laptop profile when you have time? 🙇‍♂️

deathbymanatee:
done!

aanderse:
thank you!

PR #238, opened 2026-09-03
ly: make brightnessctl and its commands optional

dmkhitaryan:


modules/services/ly/default.nix
```diff
@@ -92,6 +90,10 @@ in
         else
           (lib.getExe config.programs.xorg.package);
       xsessions = "/run/current-system/sw/share/xsessions";
+    }
+    // lib.optionalAttrs config.programs.brightnessctl.enable {
```
aanderse:
```suggestion
    // lib.optionalAttrs config.programs.brightnessctl.enable or false {
```

if you don't import the `brightnessctl` this will cause an evaluation failure

aanderse (changes requested):

modules/services/ly/default.nix
dmkhitaryan: Oh, good catch, thanks! Will fix in the morning.

dmkhitaryan: Done.

aanderse (approved):
thanks for the fix!

PR #239, opened 2026-09-03
modules/vxwm: init

willowispll:


modules/programs/vxwm/default.nix
```diff
@@ -0,0 +1,59 @@
...
+  lib,
+  ...
+}:
+let
+  cfg = config.programs.vxwm;
+
+  sessionScript = pkgs.writeShellScript "vxwm-session" ''
+    ${lib.concatStringsSep "\n" cfg.autostart}
+    exec ${cfg.package}/bin/vxwm
+  '';
```
aanderse: does `vxwm` have no hook at all to launch something? we don't do this type of thing with any of our window managers/compositors and i'd prefer to not start as it feels a bit opinionated, especially if upstream offers a solution

willowispll: no, its a fork of dwm, so, u euther have to do this or use xinitrc

willowispll: i do this because i want it to be close to the package overrides, in 1 file

PR #240, opened 2026-09-03
overlay: wip

aanderse:


dmkhitaryan:
Would it be possible to add an opt-out option for the overlay? I'm asking because having own overlay through configuring musl, some packages already include disabling systemd (along with other fixes). Would be nice to avoid duplicates!

aanderse:
> Would it be possible to add an opt-out option for the overlay?

everything as much as possible should be opt-in with `finix`, so there can be no opt-out option because that is the default :sweat_smile:

or do you mean something more specific?

dmkhitaryan:
> > Would it be possible to add an opt-out option for the overlay?
> 
> everything as much as possible should be opt-in with `finix`, so there can be no opt-out option because that is the default 😅
> 
> or do you mean something more specific?

Oh, honestly I thought this would be on by default considering it enforces disabling systemd libs on a non-systemd environment. But if not, all good.

z4il-i:
you can disable systemd dep in upower with the following:
```nix
upower = prev.upower.override { withSystemd = false; };
```

aanderse:
> ```nix
> upower = prev.upower.override { withSystemd = false; };
> ```

looks like `upower` takes `systemd` as an input, not `systemdLibs`... i wonder if that is a bug or not... :thinking:

additionally, `upower` takes both `systemd` and `udev` as inputs, so even if you set `withSystemd = false;` it won't actually disable `systemd` :upside_down_face:

presumably we could use something like `upower.override { withSystemd = false; udev = final.libudev-zero; libgudev = ... }` to actually remove the `systemd` dependency...

---

update: gets even better... the `withSystemd` flag doesn't actually do _anything_ at all, and `systemd` isn't even needed as a dep... you just need to set the unit directory, which is done unconditionally on this package - should be fixed upstream

to get rid of `systemd` in `upower` the overlay needs to do some deep overrides

z4il-i:
Presumably a bug, including the udev imports, as upower itself works fine on my gentoo system with openrc and mdevd
Also power-profiles-daemon (which also doesn't need systemd....) can be fixed with this:
```nix
  power-profiles-daemon = prev.power-profiles-daemon.overrideAttrs (old: {buildInputs = builtins.filter (p: p != prev.systemd) old.buildInputs; });
  ```
  As for libgudev, thats just a library for providing goObject bindings for udev comp apis, so we wouldn't change libgudev here (even though its needless), instead we would just 
  ```nix
    libgudev = prev.libgudev.override { udev = final.libudev-zero;  };
```

z4il-i:
```nix
upower = prev.upower.override { withIntrospection = false; withSystemd = false; libgudev = final.libgudev.override { udev = final.libudev-zero; withIntrospection = false; }; udev = final.libudev-zero; };
```
will the ``libgudev`` override also affect other things using ``libgudev``, such as ``power-profiles-daemon``?

z4il-i:
``
(prev.pythonPackagesExtensions or []) ++ [ (python-final: python-prev: { pyudev = python-prev.pyudev.override { udev = final.libudev-zero; }; })];
``
``niri`` depends on ``libinput`` which depends on ``pyudev`` which includes a udev dependency by default, replacing ``udev`` with ``lib-udevzero`` works fine, changing no functionality of either tool

PR #241, opened 2026-09-05
fix(mdevd): prevent removing valid symlinks on fast remove+add device reorder

FixeQD:
Fixes a race in `mdevd-disk.sh`'s remove handler: it deleted symlinks by name match only, without checking if the device still existed

Fast remove+add (hot-swap) could delete a symlink that was actually current


aanderse (approved):
thank you!

PR #243, opened 2026-09-06
modules/sessiond: add default reboot and poweroff commands

r0chd:


modules/services/sessiond/default.nix
```diff
@@ -58,6 +58,11 @@ in
 
     services.polkit.enable = true;
 
+    services.sessiond.settings.power = lib.mapAttrs (_: lib.mkDefault) {
```
aanderse: `finit` provides a suspend as well

modules/services/sessiond/default.nix
```diff
@@ -58,6 +58,15 @@ in
 
     services.polkit.enable = true;
 
+    services.sessiond.settings.power = {
+      reboot = lib.mkDefault [ "/run/current-system/sw/bin/reboot" ];
+      poweroff = lib.mkDefault [ "/run/current-system/sw/bin/poweroff" ];
+      suspend = lib.mkOptionDefault [
+        "${pkgs.zzz}/bin/zzz"
+        "-z"
+      ];
```
aanderse:
```suggestion
      suspend = lib.mkDefault [ "/run/current-system/sw/bin/suspend" ];
```

what about this? we would then have to adjust the `zzz` module to drop the `mkDefault`... but that should be fine

PR #248, opened 2026-09-11
modules/services/tlp: add tlp-pd

Ag1O1:
adds options and finit service for tlp-pd. power profiles daemon like dbus interface for tlp

I used the nixos module for reference and tried to make the options the same way as how other ones are defined in finix.
note that tlp-pd shouldn't be enabled with power-profiles-daemon since it uses the same dbus interface as it. The nixos module uses an assertion to stop that from happening but i didn't see those being used in finix for options that live in separate modules so i didn't add that.

### DIFF
--- a/modules/services/tlp/default.nix
+++ b/modules/services/tlp/default.nix
@@ -3,17 +3,15 @@
   lib,
   pkgs,
   ...
-}:
-let
+}: let
   cfg = config.services.tlp;
   tlpExe = lib.getExe cfg.package;
 
   format = pkgs.formats.keyValue {
-    mkKeyValue = lib.generators.mkKeyValueDefault { } "=";
+    mkKeyValue = lib.generators.mkKeyValueDefault {} "=";
     listToValue = l: "\"${toString l}\"";
   };
-in
-{
+in {
   options.services.tlp = {
     enable = lib.mkOption {
       type = lib.types.bool;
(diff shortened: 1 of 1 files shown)


modules/services/tlp/default.nix
```diff
@@ -101,6 +119,12 @@ in
       };
     };
 
+    finit.services.tlp-pd = {
+      description = "tlp-pd service";
+      command = lib.getExe cfg.pd.package;
+      runlevels = "2345";
```
aanderse:
```suggestion
```

leaving as the default is better idea here

modules/services/tlp/default.nix
```diff
@@ -32,9 +30,27 @@ in
       '';
     };
 
+    pd = {
+      enable = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = ''
+          Whether to enable tlp-pd, a power-profile-daemon like DBus interface for TLP.
+        '';
```
aanderse, line 41:
```suggestion
        description = ''
          Whether to enable [tlp-pd](${pkgs.tlp-pd.meta.homepage}) as a system service.
        '';
```

for consistency with the rest of `finix`, if you don't mind 🙇‍♂️

modules/services/tlp/default.nix
```diff
@@ -65,7 +83,7 @@ in
       };
     };
 
-    services.udev.packages = [ cfg.package ];
+    services.udev.packages = [cfg.package];
```
aanderse:
```suggestion
    services.dbus.packages = [ cfg.package ];
    services.udev.packages = [cfg.package];
```

don't we also require adding the `dbus` configuration? does this work without? 🤔

aanderse (changes requested):
looks good! just a couple minor things... also, please run the `nix fmt` check over the code!

thank you 🙇‍♂️

modules/services/tlp/default.nix
Ag1O1: Yes it works without it. But I'm not entirely sure how. Should I add it just in case?
also do you mean `services.dbus.packages = [cfg.pd.package];`

aanderse: i think you need both `services.dbus.packages = [ cfg.package ];` and `services.udev.packages = [ cfg.package ];`

how are you using this program? does it provide a command you call? or are you using something like `noctalia-shell`, which likely connects over the `dbus` interface?

Ag1O1: I am using noctalia

Ag1O1: I figured out what was happening, when tlp-pd is started it automatically registers its dbus interface. So technically adding it to dbus packages isn't strictly needed.

aanderse: i figured out the reason this happens in `nixos`/`finix`, thanks for the info!

please do add the package to `services.dbus.packages` because i plan on diverging from `nixos` and correcting this behaviour at some point

thank you!

PR #249, opened 2026-09-11
modules/sh: init

aanderse:
simplest parts of #224 extracted into a standalone PR, ish... (`coreutils` cleanup not directly related, but was fallout so made sense to take care of here as well)

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -8,6 +8,7 @@ let
       "modprobe"
       "plymouth"
       "resolvconf"
+      "sh"
       "shadow"
 
       # deprecated, remove at some point
@@ -54,6 +55,7 @@ in
       ./programs/modprobe
       ./programs/plymouth
       ./programs/resolvconf
+      ./programs/sh
       ./programs/shadow
       ./security
       ./services/dbus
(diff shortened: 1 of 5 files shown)


dmkhitaryan:
~~One issue with using `lib.getExe pkgs.someShell` is that it won't get the path for `ash` in `busybox`~~:
```Nix
nix-repl> lib.getExe pkgs.dash
"/nix/store/4rwh3g5vsgavgrd6p9658ylfzq3hqfab-dash-0.5.13.5/bin/dash"

nix-repl> lib.getExe pkgs.busybox 
"/nix/store/zb319vkbdibvy4ysgiapy2309ssz6n6d-busybox-1.37.0/bin/busybox"
```

I just tested with setting `pkgs.busybox`. It does work properly. I guess Busybox gets the correct package from the symlink? How nifty!
```bash
$(readlink /bin/sh) --list-full | grep -E '(sh|ash)$'
bin/ash
sbin/fbsplash
bin/fdflush
bin/hush
bin/sh
```

aanderse:
image

i think the `example` can adequately handle this

PR #250, opened 2026-09-12
setup-etc.sh: fix mktemp failures caused by leaked TMPDIR

FixeQD:
`TMPDIR` leaked in from before the chroot into `/mnt` so mktemp pointed at a path that didn't exist inside the chroot silently skipping the whole `/etc` symlink setup

Now unsets TMPDIR at the top of the script so mktemp uses `/tmp`


aanderse (approved):
thank you!

PR #256, opened 2026-09-12
module/dash: init

z4il-i:


aanderse:
how do other distros handle the interactive flag?

PR #257, opened 2026-09-13
fix(getty): missing ESC byte breaks color in welcome banner

FixeQD:
There was a missing `\x1b` byte before the ANSI codes so `[1;32m`/`[0m` were rendered as literal text instead of coloring the banner


aanderse:
thank you!

PR #258, opened 2026-09-13
fix(shadow): PAM hardcodes yescrypt instead of using ENCRYPT_METHOD

FixeQD:
PAM password lines for `login`/`su`/`passwd` had `yescrypt` hardcoded regardless of `ENCRYPT_METHOD` in `login.defs`, causing a mismatch and false "insecure hashing" warning on new accounts


aanderse:
amazing!

PR #259, opened 2026-09-13
fix(users): raw <N> priority prefixes leak into userborn log output

FixeQD:
userborn logs using the sd-daemon priority-prefix convention (`<6>`, `<4>`, etc.) meant for systemd-journald to strip and colorize
finix has no journald so these leaked through raw on the console

### DIFF
--- a/modules/users/default.nix
+++ b/modules/users/default.nix
@@ -57,7 +57,27 @@ in
     system.activation.scripts.users = ''
       mkdir -p /etc
 
-      ${pkgs.userborn}/bin/userborn ${configFile}
+      # userborn logs using the sd-daemon priority-prefix convention
+      # (e.g. "<6>Created group foo"), which systemd-journald normally
+      # strips and colorizes. finix has no journald, so translate the
+      # prefixes into ANSI colors ourselves instead of leaking them raw.
+      ${pkgs.userborn}/bin/userborn ${configFile} 2>&1 | while IFS= read -r line; do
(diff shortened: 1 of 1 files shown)


aanderse:
i suppose upstream `userborn` has no interest in either not doing this or optionally doing it? 

what exists in this space already? some `logger` tool can handle this already perhaps? or should we direct output to syslogs instead, which could handle this? at this point the information isn't absorbed into syslogs at all so it is lost... especially painful on systems with automatic updates...

FixeQD:
> i suppose upstream `userborn` has no interest in either not doing this or optionally doing it? 
> 
> what exists in this space already? some `logger` tool can handle this already perhaps? or should we direct output to syslogs instead, which could handle this? at this point the information isn't absorbed into syslogs at all so it is lost... especially painful on systems with automatic updates...

Yeah, I doubt upstream would want to change this, the `<N>` prefix convention is intentional on their end, it's meant for journald to consume, so from their POV it's working as designed for the systemd case. 
Not really their problem that we don't run journald

You're right that piping through `logger` is the better fix here honestly I'll rework this PR to do that instead

aanderse:
ok now you really got me thinking... can/should we just put the entire activation script through `logger`? or `tee`? it is really annoying to not have this information hit syslogs... but at the same time it is pretty useful to have it in your face as well when activating a system

and longer term... nicer to get rid of the activation scripts entirely :melting_face:

PR #260, opened 2026-09-13
fix(getty): welcome banner color, attempt #3

FixeQD:
third attempt at the same single character, pls send help. 
prev merge fixed `${"\e"}` (was literally printing "e"), then someone (me) manually typed `^[` thinking that's ESC lol it's not, it's just how terminals *display* it
It's now using `builtins.fromJSON ''"\u001b"''` so no more raw byte hand-typing, ever
tested, it's actually green now 🙏


aanderse (approved):
lol this is terrible... but i'll take it 🚀

PR #261, opened 2026-09-13
modules/nix: init

z4il-i:
Adds `programs.nix` and offloads some `services.nix-daemon` settings to `programs.nix`


aanderse:
can you please go into a bit of detail on the split here? i notice you moved `services.nix-daemon.settings.sandbox` to `programs.nix.settings` - this is an example which doesn't make sense

PR #264, opened 2026-09-16
hardware/facter: init

es-sai-fi:
Depends on: https://github.com/finix-community/finix/pull/265


modules/hardware/firmware.nix
```diff
@@ -0,0 +1,55 @@
...
+          list:
+          pkgs.buildEnv {
+            name = "firmware";
+            paths = map compressFirmware list;
+            pathsToLink = [ "/lib/firmware" ];
+            ignoreCollisions = true;
+          };
+      };
+
+      enableRedistributableFirmware = lib.mkEnableOption "firmware with a license allowing redistribution" // {
```
es-sai-fi: Should this be moved under `hardware.firmware.*`? I left it like this because this is how NixOS has it like this but we don't have to follow it too.

modules/hardware/facter/firmware.nix
```diff
@@ -0,0 +1,24 @@
...
+  inherit (config.hardware.facter) report;
+  isBaremetal = config.hardware.facter.detected.virtualisation.none.enable;
+  hasAmdCpu = facterLib.hasAmdCpu report;
+  hasIntelCpu = facterLib.hasIntelCpu report;
+in
+{
+  config = lib.mkIf (config.hardware.facter.enable && isBaremetal) {
+    # none (e.g. bare-metal)
+    # provide firmware for devices that might not have been detected by nixos-facter
+    hardware.enableRedistributableFirmware = lib.mkDefault true;
```
es-sai-fi: This will cause a lot of unused firmware to be pulled into the user's system, I don't know how other Finix users feel about this. I would expect enabling facter to handle _everything_ hardware related for me, even if it comes with some costs, but this can certainly be gated behind an optional.

modules/hardware/facter/virtualisation.nix
```diff
@@ -0,0 +1,43 @@
...
+  options,
+  ...
+}:
+let
+  inherit (config.hardware.facter) report;
+  cfg = config.hardware.facter.detected.virtualisation;
+in
+{
+  options.hardware.facter.detected.virtualisation = {
+    # TODO: idk what to do with this yet
```
es-sai-fi: We have a virtualisation qemu module, should we enable it based on this? I looked at it briefly and it had some similarities with the `config` part related to qemu in the NixOS module but I'll admit I have NO idea about this.

modules/hardware/facter/bluetooth.nix
```diff
@@ -0,0 +1,20 @@
...
+  imports = [ modules.bluetooth ];
+
+  options.hardware.facter.detected.bluetooth.enable =
+    lib.mkEnableOption "Enable the Facter bluetooth module"
+    // {
+      default = builtins.length (config.hardware.facter.report.hardware.bluetooth or [ ]) > 0;
+      defaultText = "hardware dependent";
+    };
+
+  config.services.bluetooth.enable = lib.mkIf config.hardware.facter.detected.bluetooth.enable (
```
es-sai-fi: Is this okay?

es-sai-fi:
I'll have to admit I haven't tried to run this yet, this is a first pass on trying to port facter to Finix

es-sai-fi:
Should I split "hardware: modularize firmware section, add facter" into another pr and get that merged first?

PR #265, opened 2026-09-17
hardware: modularize firmware opts, add enableRedistributableFirmware

es-sai-fi:


modules/hardware/firmware.nix
```diff
@@ -0,0 +1,74 @@
...
+      [
+        linux-firmware
+        ipw2200-firmware
+        rtl8192su-firmware
+        rt5677-firmware
+        rtl8761b-firmware
+        zd1211fw
+        alsa-firmware
+        sof-firmware
+        libreelec-dvb-firmware
```
aanderse, line 70: i feel like this belongs in a profile for home theater pcs or something instead?

aanderse (commented):
hmm... do many people need all the firmwares in this option? it does seem to have some cards which might be common enough to justify it? or are these firmware for older uncommon cards?

PR #267, opened 2026-09-18
hardware/facter: init

es-sai-fi:
Depends on: https://github.com/finix-community/finix/pull/265

I had to close last one because I fucked up :sweat:


es-sai-fi:
Works. I'll have to now look into the virtualisation part and see if it works on finix systems not running eudev/gardendevd

PR #268, opened 2026-09-18
devshell: init

es-sai-fi:


shell.nix
```diff
@@ -0,0 +1,10 @@
+let
+  pkgs = import (import ./lon.nix).nixpkgs { };
+in
+pkgs.mkShellNoCC {
+  packages = with pkgs; [
+    nil
+    nixfmt
+    statix
```
aanderse, line 7: should we include a `statix.toml`? i don't want to bikeshed on the rules for too long...

i don't mind the output if i run with this `statix.toml`:

```toml
disabled = ['repeated_keys']
nix_version = '2.4'
ignore = ['.direnv']
```

.envrc
```diff
@@ -0,0 +1 @@
+use nix
```
es-sai-fi, line 1: This is for nix2 but I did add the devshell to the flake for completenesss, should I do `use flake` here then?

shell.nix
```diff
@@ -0,0 +1,10 @@
+{
+  pkgs ? import (import ./lon.nix).nixpkgs { },
+}:
+pkgs.mkShellNoCC {
+  packages = with pkgs; [
+    nil
```
es-sai-fi: This is what I use, but there's also nixd, should I also include it or do we just rock with nil?

flake.nix
```diff
@@ -3,6 +3,16 @@
 
   outputs =
     { self }:
+    let
+      sources = import ./lon.nix;
+      lib = import (sources.nixpkgs + "/lib");
+
+      forAllSystems =
```
es-sai-fi, line 10: Let me know if you are fine with this... IMO it looks simpler like this but you have the last word.

.envrc
es-sai-fi: As much as I don't like using flakes I gotta admit it makes more sense to just use them here

shell.nix
aanderse: actually maybe we can drop ide integration, users likely provide their own choice

.envrc
aanderse: users may not have flakes enabled

flake.nix
aanderse: i think this looks really good 👍

